PR #4, opened 2017-04-26
compiler: gen server code.

disksing:
Remain 2 compile errors defer to @BusyJay 
```
$ cargo build --example server
   Compiling grpc v0.1.0 (file:///home/menglong/dev/rust/grpc-rs)
error[E0308]: mismatched types
  --> server.rs:62:93
   |
62 |     let server = route_guide_grpc::bind_route_guide(ServerBuilder::new(env), instance).bind("127.0.0.1", 50051).build();
   |                                                                                             ^^^^^^^^^^^ expected struct `std::string::String`, found reference
   |
   = note: expected type `std::string::String`
              found type `&'static str`
   = help: here are some functions which might fulfill your needs:
           - .escape_debug()
           - .escape_default()
           - .escape_unicode()
           - .to_lowercase()
           - .to_uppercase()

error: method `start` is private
  --> server.rs:63:12
   |
63 |     server.start();
   |            ^^^^^

error: aborting due to 2 previous errors

error: Could not compile `grpc`.
```

### DIFF
--- a/compiler/src/codegen.rs
+++ b/compiler/src/codegen.rs
@@ -342,8 +342,7 @@ impl<'a> ServiceGen<'a> {
 
         w.write_line("");
 
-        w.pub_fn(&*format!("bind_{}<S: {} + Send + Sync + 'static>(mut builder: \
-                          {}, s: S) -> {2}",
+        w.pub_fn(&*format!("bind_{}<S: {} + Send + 'static>(mut builder: {}, s: S) -> {2}",
                            snake_name(self.service_name()),
                            self.service_name(),
                            fq_grpc("ServerBuilder")),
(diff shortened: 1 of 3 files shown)


compiler/src/codegen.rs
```diff
@@ -244,6 +244,39 @@ impl<'a> MethodGen<'a> {
...
+    }
+
+    fn write_bind(&self, w: &mut CodeWriter) {
+        let add = match self.method_type().0 {
+            MethodType::Unary => "add_unary_handler",
+            MethodType::ClientStreaming => "add_client_streaming_handler",
+            MethodType::ServerStreaming => "add_server_streaming_handler",
+            MethodType::Dulex => "add_duplex_streaming_handler",
+        };
+        w.block(&*format!("builder = builder.{}(&{}, move |ctx, req, resp| {{",
```
overvenus: Why not `w.block(&format!())` ?

PR #24, opened 2017-05-10
reexport RpcStatusCode

BusyJay:
So user can use `RpcStatusCode` without adding grpc-sys as dependency explicitly.


ngaut:
LGTM

disksing:
LGTM.

PR #36, opened 2017-05-17
add route guide example

BusyJay:
The example route_guide_server and route_guide_client can be used with any client or server implemented in other language.


disksing:
LGTM.

PR #38, opened 2017-05-22
channel: add backoff configuration

BusyJay:


src/channel.rs
```diff
@@ -12,11 +12,12 @@
 // limitations under the License.
 
 
-use std::{mem, ptr};
+use std::{cmp, mem, ptr, usize};
```
disksing: Why usize need to be explicitly imported?

src/channel.rs
```diff
@@ -50,6 +53,11 @@ fn format_user_agent_string(agent: &str) -> CString {
     CString::new(val).unwrap()
 }
 
+fn dur_to_ms(dur: Duration) -> usize {
+    let micro = dur.as_secs() * 1000 + dur.subsec_nanos() as u64 / 1_000_000;
```
disksing: millisecond?

src/channel.rs
BusyJay: So I can use `usize::MAX`.

disksing:
LGTM.

PR #41, opened 2017-05-29
add Interoperability test suit

BusyJay:
ref: https://github.com/grpc/grpc/blob/v1.3.x/doc/interop-test-descriptions.md

### DIFF
--- a/interop/src/bin/client.rs
+++ b/interop/src/bin/client.rs
@@ -1,0 +1,111 @@
+// Copyright 2017 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+extern crate clap;
+extern crate grpc;
+extern crate grpc_proto;
+extern crate interop;
+
+use std::sync::Arc;
+
+use clap::{App, Arg};
+use grpc::{ChannelBuilder, ChannelCredentialsBuilder, Environment};
+use grpc_proto::util;
+
+use interop::Client;
+
+fn main() {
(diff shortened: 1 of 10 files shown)


BusyJay:
Compression (depends on metadata), cache (needs proxy server) and Oauth/jwt/per_rpc_creds (depends on google's service) is not implemented yet.

BusyJay:
PTAL

interop/src/bin/client.rs
```diff
@@ -0,0 +1,111 @@
...
+        "SERVER_STREAMING" => client.server_streaming(),
+        "PING_PONG" => client.ping_pong(),
+        "EMPTY_STREAM" => client.empty_stream(),
+        "CANCEL_AFTER_BEGIN" => client.cancel_after_begin(),
+        "CANCEL_AFTER_FIRST_RESPONSE" => client.cancel_after_first_response(),
+        "TIMEOUT_ON_SLEEPING_SERVER" => client.timeout_on_sleeping_server(),
+        "STATUS_CODE_AND_MESSAGE" => client.status_code_and_message(),
+        "UNIMPLEMENTED_METHOD" => client.unimplemented_method(),
+        "UNIMPLEMENTED_SERVICE" => client.unimplemented_service(),
+        _ => panic!("unknown case: {:?}", case),
```
BusyJay: I think they are the same.

interop/src/client.rs
```diff
@@ -0,0 +1,242 @@
...
+            .unwrap();
+        assert_eq!(resp_sizes, sizes);
+        println!("pass");
+    }
+
+    pub fn ping_pong(&self) {
+        print!("testing ping pong ... ");
+        let (mut sender, mut receiver) = self.client.full_duplex_call();
+        let cases = vec![(31415, 27182), (9, 8), (2653, 1828), (58979, 45904)];
+        for (resp_size, payload_size) in cases {
```
BusyJay: See L102.

overvenus: L102 only waits the stream, it does not check if the stream has exactly four responses.

BusyJay: Yes. Additional check has been added.

BusyJay:
PTAL

BusyJay:
@sre-bot try

overvenus:
LGTM

PR #67, opened 2017-07-06
bench: utilize async_(server|client)_threads

overvenus:


benchmark/src/client.rs
```diff
@@ -342,7 +342,10 @@ pub struct Client {
 
 impl Client {
     pub fn new(cfg: &ClientConfig) -> Client {
-        let env = Arc::new(EnvBuilder::new().build());
+        let mut env = EnvBuilder::new();
```
BusyJay: It's a builder not an env.

BusyJay:
LGTM

PR #96, opened 2017-10-06
Update gcc 0.3.53 to cc 1.0

illicitonion:
The crate was renamed and pushed stable.


BusyJay:
Thanks, but I think it's duplicated with #94.

illicitonion:
Indeed it is :) Thanks!

PR #117, opened 2017-11-17
Add logger for benchmark

overvenus:


benchmark/src/main.rs
```diff
@@ -42,6 +45,7 @@ fn main() {
         .get_matches();
     let port: u16 = matches.value_of("port").unwrap_or("8080").parse().unwrap();
 
+    let _log_guard = init_log(env::var(LOG_FILE).ok());
```
BusyJay: It should print to stderr by default.

BusyJay:
LGTM

PR #127, opened 2017-12-11
benchmark: fix shutdown failure

BusyJay:
This pr fixes an hang-up-forever issue during benchmark.

To add all the stats 1.7.2 introduces, we need a public API and compatible rust-protobuf. See grpc/grpc#13705 and stepancheg/rust-protobuf#260.

/cc #107 #118 

After this pr is merged, I think we are ready to release 0.2.0.

### DIFF
--- a/benchmark/src/bench.rs
+++ b/benchmark/src/bench.rs
@@ -12,6 +12,10 @@
 // limitations under the License.
 
 
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+
 use grpc_proto::testing::services_grpc::BenchmarkService;
 use grpc_proto::testing::messages::{SimpleRequest, SimpleResponse};
 use grpc_proto::util;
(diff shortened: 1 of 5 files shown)


benchmark/src/client.rs
```diff
@@ -143,6 +143,17 @@ impl<B: Backoff> ExecutorContext<B> {
     }
 }
 
+// Since impl trait is not stable yet, implement this as a function is impossible without box.
+macro_rules! spawn {
```
overvenus: What's the difference between the `fn spawn()` above?

BusyJay: `client` can be different types.

disksing:
PTAL @AndreMouche

benchmark/src/bench.rs
```diff
@@ -26,16 +30,28 @@ fn gen_resp(req: SimpleRequest) -> SimpleResponse {
     resp
 }
 
+
```
AndreMouche: redundant empty line?

benchmark/src/client.rs
AndreMouche: Could we remove the above `fn spawn()` ?

BusyJay:
ping

PR #134, opened 2017-12-21
*: support zero copy

BusyJay:
In the past, deserialization introduces two times of copy. First is loading the message payload from gRPC core buffer to rust wrapper buffer, the second time is deserialization in codec, which only accepts a slice of bytes as parameter.

This pr exposes the underlying byte buffer as a zero copy reader: `MessageReader`. So implementer can choose whether to copy or not.

### DIFF
--- a/grpc-sys/src/lib.rs
+++ b/grpc-sys/src/lib.rs
@@ -198,7 +198,18 @@ pub enum GrpcBatchContext {}
 pub enum GrpcServer {}
 pub enum GrpcRequestCallContext {}
 pub enum GrpcAlarm {}
-pub enum GrpcWrapByteBufferReader {}
+
+#[repr(C)]
+pub union GrpcByteBufferReaderCurrent {
+    pub index: c_uint,
+}
+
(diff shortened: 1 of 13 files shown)


grpc-sys/src/lib.rs
```diff
@@ -199,6 +199,53 @@ pub enum GrpcServer {}
...
+
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct GrpcSlice {
+    ref_count: *mut GrpcSliceRefCount,
+    data: GrpcSliceData,
+}
+
+#[repr(C)]
+#[repr(C)]
```
ice1000: Why two `#[repr(c)]` here?

BusyJay: It should be an unexpected change when tweaking around.

Hoverbear:
With #222 and #225 should we close this?

ice1000:
@Hoverbear Correction: now it's #238 .

PR #135, opened 2017-12-21
Custom metadata functionality [WIP]

functor:
It would be useful to include custom metadata functionality. 

Currently, it is not implemented, but some of the needed hooks are already there.

Given the current code, one approach that seems relatively straightforward would be to add a metadata array option to `CallOption` so that clients can supply a custom metadata using the `_opt` generated methods. 

On the server/service side, a possible approach is to add metadata accessor via `RpcContext` so that service implementation can query the custom metadata.

A minimal expanded hello_world example of this can be seen at /examples/hello_world_metadata/ [client.rs](https://github.com/functor/grpc-rs/blob/feature/metadata/examples/hello_world_metadata/client.rs) and [server.rs](https://github.com/functor/grpc-rs/blob/feature/metadata/examples/hello_world_metadata/server.rs)

I'm would be happy to help with this feature. This PR points to the relevant work-in-progress fork.

It needs more work before merging, but suggestions at this stage are welcomed.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -3,15 +3,19 @@ use std::ffi::CString;
 use libc::c_char;
 use std::ascii::AsciiExt;
 
+/// Builder used to construct a MetadataArray value.
 pub struct MetadataArrayBuilder {
     entries: Vec<(*mut c_char, *mut c_char, usize)>
 }
 
 impl MetadataArrayBuilder {
+
+    /// Create a new empty builder.
     pub fn new() -> MetadataArrayBuilder {
         MetadataArrayBuilder { entries: vec![] }
     }
 
+    /// Add a new key-value pair to the metadata being built.
     pub fn add(mut self, key: Vec<u8>, value: Vec<u8>) -> MetadataArrayBuilder {
         assert!(key.iter()
             .all(|b|
@@ -35,6 +39,7 @@ impl MetadataArrayBuilder {
         self
     }
 
+    /// Build the metadata array data type that can be used in CallOption
     pub fn build(self) -> MetadataArray {
         let array_size = self.entries.len();
         let array = unsafe { grpc_sys::grpcwrap_metadata_array_create(array_size) };
@@ -47,6 +52,7 @@ impl MetadataArrayBuilder {
     }
 }
 
+/// Metadata data type used in CallOption datatype.
 pub struct MetadataArray {
     array: *mut GrpcMetadataArray,
 }
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -1,6 +1,7 @@
 use grpc_sys::{self, GrpcMetadataArray};
 use std::ffi::CString;
 use libc::c_char;
+use std::ascii::AsciiExt;
 
 pub struct MetadataArrayBuilder {
     entries: Vec<(*mut c_char, *mut c_char, usize)>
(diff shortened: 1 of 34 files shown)


src/metadata.rs
```diff
@@ -0,0 +1,109 @@
+use grpc_sys::{self, GrpcMetadataArray};
+use std::ffi::CString;
+use libc::c_char;
+#[allow(unused)]
+use std::ascii::AsciiExt;
+use std::slice;
+
+/// Builder used to construct a MetadataArray value.
+pub struct MetadataArrayBuilder {
+    entries: Vec<(*mut c_char, *mut c_char, usize)>
```
BusyJay: Memory leak if `build` is not called at the end. Why not keep it as `CString`?

src/metadata.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Drop for MetadataArray {
+    fn drop(&mut self) {
+        unsafe { grpc_sys::grpcwrap_metadata_array_destroy_full(self.array) }
+    }
+}
+
+/// Immutable handle to the core MetadataArray data type with accessor methods.
+pub struct MetadataArrayView {
```
BusyJay: I think just `MetadataArray` is enough. You can make `as_mut_ptr` as `pub(crate)`.

src/metadata.rs
functor: Good point. Maybe it's better to represent it as Vec<(CString, Vec<u8>). The key is expected to be a null terminated c-style string so CString seems appropriate for that, while the value is treated as an opaque byte array so better to keep it as Vec<u8>. 

[relevant commit](https://github.com/pingcap/grpc-rs/pull/135/commits/345821e1365ace32d0df70242d3a63bc21ae7eae)

src/metadata.rs
functor: I agree, and this was just a rough first pass at it. It would be better to avoid having these two data types `MetadataArray` and `MetadataArrayView`.

However there might be a bit more to it than controlling visibility of as_mut_ptr.

Currently `MetadataArray` is used to handle the mutable GrpcMetadataArray used in `grpcwrap_call_X` calls. So it needs to be mutable because `grpcwrap_call_X` call `grpcwrap_metadata_array_move` which modifies the struct. However, the struct itself needs to be deallocated on the rust side (done via `Drop`.) 

In contrast `MetadataArrayView` is used for a `*const GrpcMetadataArray` returned from `grpcwrap_request_call_context_metadata`, and it is not supposed to be deallocated on the rust side, it is managed on the c side. Because of that, its lifetime should be dealt with appropriately so that it can't be used to reference the underlying memory that might be no longer invalid. That probably means that it should have the lifetime of corresponding `RequestContext` etc. but need to look into this.

But yes, it would be nice to unify these two ...

BusyJay: Actually it's just the `Owned` and `Borrowed` form of the same struct/data. So you can pass `MetadataArray` to client call and return `&MetadataArray` from server call.

functor: Maybe we're not talking about the same thing? So who would own the `&MetadataArray` on the server call? Currently the server call provides `&MetadataArrayView` (with the same lifetime as the corresponding `RpcContext`) via the `RpcContext::metadata()` method. When the `MetadataArrayView` is deallocated (with the`RpcContext`) the underlying `*const GrpcMetadataArray` is *not deallocated*.

If `&MetadataArray` was used instead (as it is implemented now,) then the corresponding `*const GrpcMetadataArray` *would be deallocated*, which would be a double deallocate error, since on the server call this memory is all managed by grpc-core.

BusyJay: > ... So who would own the &MetadataArray on the server call? ...

`RpcContext` will own the `MetadataArray` and `&MetadataArray` borrows it.

> If &MetadataArray was used instead (as it is implemented now,) then the corresponding *const GrpcMetadataArray would be deallocated, ...

No, it shouldn't. Because only a reference is returned. So just the reference is deallocated rather than the data it borrows.

I think the problem here is `GrpcMetadataArray` is defined as an `enum`. If it's declared as a C struct, then `*const GrpcMetadataArray` is free and safe to be cast to `&GrpcMetadataArray`.

BusyJay:
@functor Hi, any progress on this? Is there anything I can help?

functor:
@BusyJay Hi, I haven't spent time on it, but I should be able to revisit it in the next couple of days.

src/metadata.rs
functor: I have trouble understanding your previous suggestions. Perhaps you could take a stab at modifying the code to safely unify MetadataArray and MetadataArrayView? (You should be able to push into feature/metadata.)

BusyJay:
Hi, I have implemented this and sent a new pr as #161. Feel free to leave a comment if you find any problem, thank you.

PR #140, opened 2018-01-04
CallOptions are passed by reference

illicitonion:
This allows them to be re-used for multiple calls, rather than requiring
frequent re-construction.


BusyJay:
Maybe `Clone` can be implemented for `CallOption` if caller wants to use the same option for all calls.

BusyJay:
Fixed by #143.

PR #141, opened 2018-01-04
CallOption is Copy and Clone

illicitonion:


BusyJay:
It's not anymore when #135 is merged.

BusyJay:
Duplicated with #143

PR #145, opened 2018-01-06
Added support for Google default credentials.

Snnappie:
This change adds support for authenticating with [Google Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials).

This is necessary to connect to Google APIs via gRPC.

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -11,7 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
 use std::ffi::CString;
 use std::ptr;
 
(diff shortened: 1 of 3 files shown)


src/credentials.rs
```diff
@@ -200,6 +199,23 @@ impl ChannelCredentials {
     pub fn as_mut_ptr(&mut self) -> *mut GrpcChannelCredentials {
         self.creds
     }
+
+    /// Attempts to construct a `ChannelCredentials` that is authenticated with
+    /// Google OAuth credentials.
+    pub fn google_default_credentials() -> Result<ChannelCredentials, String> {
```
BusyJay: Please use the error from `src/error.rs`  instead.

Snnappie: @BusyJay Done - Let me know if there is anything else. Thanks

BusyJay:
LGTM

BusyJay:
Thanks for your contribution!

PR #147, opened 2018-01-08
Specifying gRPC event engine

overvenus:


BusyJay:
I don't think it's a good idea. I prefer keeping engine selections inside `TiKV` project and let the library has the same default engine as gRPC core. Besides, `init_grpc` is not just be called inside `EnvBuilder` after #145 is merged.

overvenus:
It's just an option and linux only, it will not change the default engine.

PR #161, opened 2018-02-09
*: implement request header support

BusyJay:
This pr adds support to customize request headers. We need it to examine cluster id in TiKV.

close #135

### DIFF
--- a/grpc-sys/grpc_wrap.c
+++ b/grpc-sys/grpc_wrap.c
@@ -164,12 +164,17 @@ grpcwrap_metadata_array_destroy_full(grpc_metadata_array *array) {
  */
 GPR_EXPORT void GPR_CALLTYPE
 grpcwrap_metadata_array_init(grpc_metadata_array *array, size_t capacity) {
-  if (capacity) {
-    grpc_metadata *arr =
-        (grpc_metadata *)gpr_malloc(sizeof(grpc_metadata) * capacity);
-    memset(arr, 0, sizeof(grpc_metadata) * capacity);
-    array->metadata = arr;
+  array->count = 0;
+  array->capacity = capacity;
+  if (!capacity) {
+    array->metadata = NULL;
+    return;
   }
+
+  grpc_metadata *arr =
+      (grpc_metadata *)gpr_malloc(sizeof(grpc_metadata) * capacity);
+  memset(arr, 0, sizeof(grpc_metadata) * capacity);
+  array->metadata = arr;
 }
 
 GPR_EXPORT void GPR_CALLTYPE grpcwrap_metadata_array_add(
@@ -208,7 +213,7 @@ grpcwrap_metadata_array_cleanup(grpc_metadata_array *array) {
 }
 
 GPR_EXPORT void GPR_CALLTYPE
-grpcwrap_metadata_array_freeze(grpc_metadata_array *array) {
+grpcwrap_metadata_array_shrink_to_fit(grpc_metadata_array *array) {
   if (array->count == array->capacity) {
     return;
   }
--- a/grpc-sys/grpc_wrap.c
+++ b/grpc-sys/grpc_wrap.c
@@ -160,53 +160,63 @@ grpcwrap_metadata_array_destroy_full(grpc_metadata_array *array) {
 }
 
 /*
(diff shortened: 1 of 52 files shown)


grpc-sys/grpc_wrap.c
```diff
@@ -160,55 +160,69 @@ grpcwrap_metadata_array_destroy_full(grpc_metadata_array *array) {
...
-grpcwrap_metadata_array_create(size_t capacity) {
-  grpc_metadata_array *array =
-      (grpc_metadata_array *)gpr_malloc(sizeof(grpc_metadata_array));
-  grpc_metadata_array_init(array);
-  array->capacity = capacity;
-  array->count = 0;
-  if (capacity > 0) {
-    array->metadata =
+GPR_EXPORT void GPR_CALLTYPE
+grpcwrap_metadata_array_init(grpc_metadata_array *array, size_t capacity) {
```
huachaohuang: Should init `count` and `capacity` properly too.

BusyJay:
PTAL

grpc-sys/grpc_wrap.c
```diff
@@ -160,55 +160,69 @@ grpcwrap_metadata_array_destroy_full(grpc_metadata_array *array) {
...
   return (char *)GRPC_SLICE_START_PTR(array->metadata[index].value);
 }
 
+GPR_EXPORT void GPR_CALLTYPE
+grpcwrap_metadata_array_cleanup(grpc_metadata_array *array) {
+  grpcwrap_metadata_array_destroy_metadata_including_entries(array);
+}
+
+GPR_EXPORT void GPR_CALLTYPE
+grpcwrap_metadata_array_freeze(grpc_metadata_array *array) {
```
huachaohuang: `freeze` seems not intuitive, consider using `shrink_to_fit` from C++ STL.

src/metadata.rs
```diff
@@ -0,0 +1,331 @@
...
+
+use grpc_sys::{self, GrpcMetadataArray};
+use std::{ptr, slice, str};
+use std::borrow::Cow;
+
+use libc;
+
+use error::{Error, Result};
+
+fn normalize_key(key: &str, binary: bool) -> Result<Cow<str>> {
```
huachaohuang: Is this required by gRPC? Any reference?

src/metadata.rs
```diff
@@ -0,0 +1,331 @@
...
+///
+/// - Response trailers
+///
+///     They are sent by the server at the end of a remote call along with
+///     resulting call status.
+///
+/// Metadata value can be ascii string or bytes. They are distinguish by the
+/// key suffix, key of bytes value should have suffix '-bin'.
+#[repr(C)]
+pub struct Metadata(GrpcMetadataArray);
```
huachaohuang: I think it is better to name it `MetadataArray` to keep the naming consistency with gRPC, it is weird that `Metadata` actually contains an array of metadata.

src/metadata.rs
BusyJay: See docs at https://github.com/pingcap/grpc-rs/pull/161/files#diff-26aa3c8895057aa93d60e92c0dee6ddcR145

src/metadata.rs
BusyJay: Actually its name is taken from [gRPC cshar](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/Metadata.cs#L37).

tests/cases/metadata.rs
```diff
@@ -0,0 +1,80 @@
...
+        .unwrap()
+        .add_bytes("k1-bin", &[0x00, 0x01, 0x02])
+        .unwrap();
+    let metadata = builder.build();
+    let call_opt = CallOption::default().headers(metadata);
+
+    let mut req = HelloRequest::new();
+    req.set_name("world".to_owned());
+    let resp = client
+        .say_hello_async_opt(&req, call_opt)
```
huachaohuang: Why not just `say_hello_opt`?

BusyJay:
@siddontang @overvenus @huachaohuang PTAL

huachaohuang:
LGTM

BusyJay:
PTAL

PR #176, opened 2018-04-13
Update dependencies

Hoverbear:
Update various dependencies.


Hoverbear:
So I originally held this because of Futures 0.2, however https://github.com/pingcap/grpc-rs/issues/174#issuecomment-380559207 suggests it might be worth just waiting for 0.3.

I would put slog in another PR.

BusyJay:
LGTM, @overvenus PTAL

PR #185, opened 2018-05-14
channel: fix overflow

BusyJay:
usize is too large for c int. This pr changes the type from usize to i32, which suits most modern compiler and architect. Also add an assert check that prevent overflow from happening at run time for the others.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -395,7 +395,7 @@ impl ChannelBuilder {
             let key = k.as_ptr() as *const c_char;
             match *v {
                 Options::Integer(val) => unsafe {
-                    // On most morden compiler and architect, c_int is the same as i32,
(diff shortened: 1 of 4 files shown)


src/channel.rs
```diff
@@ -396,6 +395,14 @@ impl ChannelBuilder {
             let key = k.as_ptr() as *const c_char;
             match *v {
                 Options::Integer(val) => unsafe {
+                    // On most morden compiler and architect, c_int is the same as i32,
```
Hoverbear: `modern`

huachaohuang:
LGTM

BusyJay:
PTAL

BusyJay:
Touch a commit to trigger appveyor, PTAL

Issue #183, opened 2018-05-13
protobuf dependency is really outdated


Kane-Sendgrid:
Current protobuf dependency is 1.2 a year old, latest is 1.6.
Because of that:
```
cargo install grpcio-compiler 
```
Is failing, it's trying to download protobuf 1.6.

thrawn01:
See PR #182

Hoverbear:
This has been fixed! :)

PR #187, opened 2018-05-17
Use protobuf ~2.0

Snnappie:
protobuf 1.6 had breaking changes, which was incorrect according to semver. This was correct by publishing 1.5 as 1.7.1, and 1.6 as 2.0. But since grpc-rs was updated to support the 1.6 changes, it may as well be moved forward to protobuf 2.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.2.2"
+version = "0.2.3"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
 keywords = ["grpc", "protobuf", "rpc", "tls", "http2"]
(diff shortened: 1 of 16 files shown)


Hoverbear:
Also related to #183

Cargo.toml
```diff
@@ -18,7 +18,7 @@ all-features = true
 grpcio-sys = { path = "grpc-sys", version = "0.2.1" }
 libc = "0.2"
 futures = "^0.1.15"
-protobuf = { version = "1.6", optional = true }
+protobuf = { version = "~2.0", optional = true }
```
BusyJay: Why add `~`?

Snnappie: `~` limits the range to `>= 2.0.0 < 2.1.0`. `2.0` is the same as `^2.0`, which is `>=2.0.0 < 3.0.0`. Seemed like the smaller version range was preferable to prevent this from being an issue again. I can change it back to just `= "2.0"` if you would prefer

BusyJay:
Thanks for the contribution.

PR #201, opened 2018-07-04
build: use fake go when secure feature is disabled

BusyJay:
Address the issue mentioned in #199.

All CIs pre-install go by default, and there seems to be no configuration entries available to tweak the behavior. So I leave the tests as it is.


BusyJay:
PTAL

PR #207, opened 2018-07-13
Reduce warning and remove redundant variables

iamcharleschege:
Removed redundant variables as pointed out by clippy and some files that might have caused build to fail on travis ci


BusyJay:
Thanks! LGTM, @overvenus PTAL.

PR #215, opened 2018-08-08
*: format code with latest stable rustfmt

BusyJay:
Keep compatible with rustc 1.28.0, which is the latest stable rustc.


Cargo.toml
```diff
@@ -10,6 +10,7 @@ homepage = "https://github.com/pingcap/grpc-rs"
 documentation = "https://docs.rs/grpcio"
 description = "The rust language implementation of gRPC, base on the gRPC c core library."
 categories = ["asynchronous", "network-programming"]
+autoexamples = false
```
Hoverbear: Seems to be an unrelated change?

BusyJay: I change the pr title to better reflect the changes.

overvenus:
Please fix confliction.

Cargo.toml
BusyJay: It's possible to let it autodiscover, which requires changes to examples' structure. Set it to false is the simplest way to fix the warnings.

BusyJay:
PTAL

BusyJay:
I fix the warnings reported by clippy and then always enable clippy check now. PTAL

src/call/mod.rs
```diff
@@ -541,12 +534,12 @@ impl WriteFlags {
     }
 
     /// Get whether buffer hint is enabled.
-    pub fn get_buffer_hint(&self) -> bool {
+    pub fn get_buffer_hint(self) -> bool {
```
Hoverbear: This seems to be an API change. Is this intentional? (also the below function)

They seem to now consume the `self`, which seems unnecessary for ` `get_*` check.

src/call/mod.rs
```diff
@@ -541,12 +534,12 @@ impl WriteFlags {
...
         (self.flags & grpc_sys::GRPC_WRITE_BUFFER_HINT) != 0
     }
 
     /// Get whether compression is disabled.
-    pub fn get_force_no_compress(&self) -> bool {
+    pub fn get_force_no_compress(self) -> bool {
```
Hoverbear: This one also consumes now.

src/call/mod.rs
BusyJay: `self` is `Copy`, so even it's an API change, it's backward compatible.

src/call/mod.rs
BusyJay: The same as above.

PR #222, opened 2018-08-25
Re-work on zero-copy

ice1000:
This is not ready to merge yet, but I'm opening this PR so we can discuss changes related to #134 .


grpc-sys/src/lib.rs
```diff
@@ -348,6 +349,53 @@ pub struct GrpcMetadataArray {
...
+
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct GrpcSlice {
+    ref_count: *mut GrpcSliceRefCount,
+    data: GrpcSliceData,
+}
+
+#[repr(C)]
+#[repr(C)]
```
ice1000: This line was added by @BusyJay in #134 , I don't understand why two `#[repr(c)]` but I kept it.

ice1000:
I think the use of `Option<MessageReader>` in #134 should be eliminated. The only usage of passing `None` is

https://github.com/pingcap/grpc-rs/blob/8fff94905ca9bccc1875f7f5a9cae7d12aa17196/src/call/server.rs#L79

. Seems that we need some kind of `MessageReader` that reads nothing.

BusyJay:
Why send another pull request? If you are going to "rework" somebody's pull request, better talk about it with the author first.

ice1000:
@BusyJay Because I cannot find you on Slack when I started work on this ([I tried](https://pingcap.slack.com/archives/C0ZCYECU8/p1534645139000100) to, [not once](https://pingcap.slack.com/archives/C0ZCYECU8/p1534701121000200), because as you see I have some [questions](https://github.com/pingcap/grpc-rs/pull/134#discussion_r212803953) about your code).
@overvenus said you rarely respond to people online.

I opened this new PR because I consider #134 too old and it's completely diverse with the current master branch.

ice1000:
@BusyJay If you have any thoughts about this PR, don't hesitate to tell me now.

BusyJay:
I'm not always online on weekend, better contact me on weekdays if you expect an instant response.

ice1000:
> If you are going to "rework" somebody's pull request, better talk about it with the author first.

Now I'm talking to you. Do you have something to say?

ice1000:
I have another question: none of the implemented functions of `Read`/`BufRead` for `MessageReader` are returning `Err`s. (Some seems to, say, `read_to_end` returns `Err` only when `read` returns `Err`, `read` returns `Err` only when `fill_buf` returns `Err`. I assumed these functions not to returning `Err` at all and you wrote them like that only to fit the interface's functions' type signatures. Then I provided `*_directly` versions for them. Is this proper? Have I missed anything? The old functions are not removed but changed implementations so there's backward compatibility.

https://github.com/pingcap/grpc-rs/pull/222/commits/c9ae7e9c3385a3d2843cedd629523a11eade3003

BusyJay:
`Read` and `BufRead` are traits from std library. `MessageReader` are only used to support serialization/deserialization, where both traits are commonly used. Implementing them by default makes it easier for people to write their own ser/de hooks.

ice1000:
This is basically done. PTAL @BusyJay @overvenus

src/call/client.rs
```diff
@@ -400,12 +406,12 @@ impl<H: ShareCallHolder, T> ResponseStreamImpl<H, T> {
...
         })?;
 
-        let mut bytes = None;
+        let mut bytes = MessageReader::default();
         loop {
             if !self.read_done {
                 if let Some(ref mut msg_f) = self.msg_f {
                     bytes = try_ready!(msg_f.poll());
-                    if bytes.is_none() {
+                    if bytes.empty {
```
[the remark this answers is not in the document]
BusyJay: Original code makes more sense to me. `Option<MessageReader>` is not the same as `MessageReader`. When `Option<MessageReader>` is `None`, it's not supposed to be read in the first place. Current implementation is easier to write bugs.

src/call/mod.rs
```diff
@@ -108,6 +110,142 @@ impl RpcStatus {
...
+pub struct MessageReader {
+    buf: *mut GrpcByteBuffer,
+    reader: GrpcByteBufferReader,
+    slice: Option<GrpcSlice>,
+    // Hack: apparently its lifetime depends on `self.slice`. However
+    // it's not going to be accessed by others directly, hence it's safe
+    // to mark it static here.
+    bytes: &'static [u8],
+    length: usize,
+    empty: bool,
```
BusyJay: This is confusing. As a reader, it should only have two states, like EOF is not reached and EOF is reached. As stated above, `Option<MessageReader>` is not the same as `MessageReader`, `None` is not empty, it means the reader doesn't exist at all. Let's make it clear.

src/call/client.rs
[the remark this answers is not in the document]
ice1000: @overvenus No, I've added a bool to mark whether it's created as `default`, so `MessageReader::default()` is different from a zero-sized result of `recv_message`

src/call/mod.rs
ice1000: `empty` here takes the place of `Option`. Now I'm somewhat changed my mind -- `Option` is now good to me.

ice1000:
Fixed, PTAL @BusyJay @overvenus

src/call/server.rs
```diff
@@ -204,10 +204,15 @@ impl UnaryRequestContext {
...
+        mut self,
+        rc: &RequestCallContext,
+        cq: &CompletionQueue,
+        reader: Option<MessageReader>
+    ) {
         let handler = unsafe { rc.get_handler(self.request.method()).unwrap() };
-        if let Some(data) = data {
-            return execute(self.request, cq, data, handler);
+        if let Some(reader) = reader {
+            return execute(self.request, cq, Some(reader), handler);
```
ice1000: Why not passing `reader` to `execute` directly?

ice1000: Oh, I see. Nevermind.

ice1000:
@siddontang no. I thought there can be some improvements but now not. And PTAL my questions above.

src/call/mod.rs
```diff
@@ -146,23 +268,28 @@ impl BatchContext {
...
-        if len == usize::MAX {
-            return None;
-        }
-        let mut buffer = Vec::with_capacity(len);
+    pub fn recv_message(&self) -> Option<MessageReader> {
+        let buf;
+        let mut reader;
+        let length;
         unsafe {
-            grpc_sys::grpcwrap_batch_context_recv_message_to_buffer(
```
ice1000: (0): here we called a C function that does `memcpy`s inside, which copies all the received message into `buffer`. Here's the only copy that happens in the original version, it's an in place copy.

src/call/mod.rs
```diff
@@ -108,6 +110,126 @@ impl RpcStatus {
...
+        let amt = {
+            let bytes = self.fill_buf_directly();
+            if bytes.is_empty() {
+                return 0;
+            }
+            let amt = cmp::min(buf.len(), bytes.len());
+            if amt == 1 {
+                buf[0] = bytes[0];
+            } else {
+                buf[..amt].copy_from_slice(bytes);
```
ice1000: (1): But when it gets the slice, it still copies.

ice1000:
The ideal approach is to use the C++ memory directly, which is not present in #134 . Am I expected to achieve that, or keep the current implementation? (Cus it can be lazily read now and it's better than in-place copy)

ice1000:
And since the destruction of `MessageReader` is done in `Drop::drop`, does that mean we can actually read the data without copying if we only call `fill_buf_directly` without `read_directly` or `read_to_end_directly`?

ice1000:
Information get. That looks good

src/call/mod.rs
```diff
@@ -108,6 +110,126 @@ impl RpcStatus {
...
+                debug_assert!(code != 0);
+                let ptr = grpc_sys::grpcwrap_slice_raw(slice, &mut len);
+                self.bytes = slice::from_raw_parts(ptr as _, len);
+            }
+        }
+        self.bytes
+    }
+}
+
+unsafe impl Sync for MessageReader {}
```
siddontang: em, strange to support Sync and Send for a Reader?

Can this Reader be used crossing threads?

ice1000: This is copied from #134 , I thought it's necessary for something else so I didn't delete it.

src/call/mod.rs
```diff
@@ -386,12 +397,15 @@ impl Call {
     ) -> Result<BatchFuture> {
         let _cq_ref = self.cq.borrow()?;
         let i = if initial_meta { 1 } else { 0 };
-        let f = check_run(BatchType::Finish, |ctx, tag| unsafe {
+        let f = check_run_with_content(BatchType::Finish, Err(Vec::from(msg)), |content,
```
ice1000: A copy is present here, this replaces another copy in C side. We can think of a way to remove this copy to achieve real zero-copy.

IMO removing the copy needs the moving `msg`, but currently, we're **borrowing**. Any chance to change the API?

grpc-sys/grpc_wrap.c
```diff
@@ -76,10 +76,17 @@
 #define GPR_CALLTYPE
 #endif
 
+grpc_slice grpc_slice_from_buffer(const char *source, size_t length) {
+    if (length == 0) return grpc_empty_slice();
+    grpc_slice slice = grpc_slice_malloc_large(length);
```
overvenus: why do we need to malloc?

ice1000: gRPC's slice uses either rc gc or copy semantics, I was still investigating about how to remove this malloc (cus this malloc the memories for the rc functions' vtable as well, I can create a fake rc to replace the built-in one)

src/call/mod.rs
```diff
@@ -108,18 +111,147 @@ impl RpcStatus {
...
+            grpc_sys::grpc_byte_buffer_reader_destroy(&mut self.reader);
+            if let Some(slice) = self.slice {
+                grpc_sys::grpc_slice_unref(slice);
+            }
+            grpc_sys::grpc_byte_buffer_destroy(self.buf);
+        }
+    }
+}
+
+type Either<A, B> = result::Result<A, B>;
```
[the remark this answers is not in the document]
ice1000: You mean I'm free to add a dependency? This is internal representation so I thought it's unnecessary, but I don't mind using a better `either`

src/call/mod.rs
```diff
@@ -108,18 +111,147 @@ impl RpcStatus {
...
+
 /// Context for batch request.
 pub struct BatchContext {
     ctx: *mut GrpcBatchContext,
+    content: Either<Box<[u8]>, Vec<u8>>,
```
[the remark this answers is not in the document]
ice1000: Sure

src/call/mod.rs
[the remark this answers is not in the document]
kennytm: I find it strange you need an `Either<Box<[u8]>, Vec<u8>>`.

A `Box<[T]>` can be freely upgraded to a `Vec<T>` with [`.into_vec()`](https://doc.rust-lang.org/std/primitive.slice.html#method.into_vec), and a `Vec<T>` can be converted back to a `Box<[T]>` at the cost of a realloc() with [`.into_boxed_slice()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice). So depending on usage you could simply stick to one of the two types.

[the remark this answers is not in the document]
ice1000: Ok. Will fix now.

ice1000:
@kennytm https://travis-ci.org/pingcap/grpc-rs/jobs/423982678#L586-L626

kennytm:
@ice1000 The `#![allow(renamed_and_removed_lints)]` needs to be added to [`src/lib.rs`](https://github.com/pingcap/grpc-rs/blob/master/src/lib.rs) as well. The two `lib.rs` belongs to different crates, and their lint settings won't affect each other.

ice1000:
image

Why it's saying test failed...

ice1000:
The merge of this issue can close #134 .

ice1000:
image
This shouldn't be my fault..

src/async/promise.rs
```diff
@@ -69,7 +69,7 @@ impl Batch {
             if succeed {
                 let status = self.ctx.rpc_status();
                 if status.status == RpcStatusCode::Ok {
-                    guard.set_result(Ok(None))
+                    guard.set_result(Ok(Default::default()))
```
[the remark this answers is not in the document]
ice1000: This is a leftover of a previous refactoring. Should I revert this?

src/call/mod.rs
```diff
@@ -108,6 +110,126 @@ impl RpcStatus {
...
+}
+
+impl MessageReader {
+    /// Get the available bytes count of the reader.
+    #[inline]
+    pub fn pending_bytes_count(&self) -> usize {
+        self.length
+    }
+
+    pub fn read_directly(&mut self, buf: &mut [u8]) -> usize {
```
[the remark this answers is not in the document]
ice1000: This one does not need you to `.unwrap()`, since it's logically safe.

grpc-sys/src/lib.rs
```diff
@@ -348,20 +350,65 @@ pub struct GrpcMetadataArray {
...
+}
+
+#[repr(C)]
+pub struct GrpcByteBufferReader {
+    pub buffer_in: *mut GrpcByteBuffer,
+    pub buffer_out: *mut GrpcByteBuffer,
+    current: GrpcByteBufferReaderCurrent,
+}
+
+#[cfg_attr(feature = "cargo-clippy", allow(clippy::unreadable_literal))]
```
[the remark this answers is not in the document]
ice1000: This isn't working.

grpc-sys/src/lib.rs
```diff
@@ -449,17 +497,35 @@ extern "C" {
...
+    pub fn grpc_byte_buffer_reader_destroy(reader: *mut GrpcByteBufferReader);
+    pub fn grpc_byte_buffer_destroy(buf: *mut GrpcByteBuffer);
+
     pub fn grpcwrap_batch_context_create() -> *mut GrpcBatchContext;
     pub fn grpcwrap_batch_context_destroy(ctx: *mut GrpcBatchContext);
     pub fn grpcwrap_batch_context_recv_initial_metadata(
         ctx: *mut GrpcBatchContext,
     ) -> *const GrpcMetadataArray;
     pub fn grpcwrap_batch_context_recv_message_length(ctx: *mut GrpcBatchContext) -> size_t;
+    #[deprecated(note="this is copying the original data")]
```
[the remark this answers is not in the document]
ice1000: Ok.

src/call/mod.rs
```diff
@@ -146,23 +268,28 @@ impl BatchContext {
...
     /// Fetch the response bytes of the rpc call.
-    // TODO: return Read instead.
-    pub fn recv_message(&self) -> Option<Vec<u8>> {
-        // TODO: avoid copy
-        let len = unsafe { grpc_sys::grpcwrap_batch_context_recv_message_length(self.ctx) };
-        if len == usize::MAX {
-            return None;
-        }
-        let mut buffer = Vec::with_capacity(len);
+    pub fn recv_message(&self) -> Option<MessageReader> {
```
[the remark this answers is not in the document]
ice1000: I'm not doing this in this pr. This can be easily fixed in #225 but very difficult in this pr.

grpc-sys/src/lib.rs
```diff
@@ -348,20 +351,65 @@ pub struct GrpcMetadataArray {
...
+#[derive(Clone, Copy)]
+pub union GrpcSliceData {
+    ref_counted: GrpcSliceRefCounted,
+    inlined: GrpcSliceInlined,
+}
+
+pub enum GrpcSliceRefCount {}
+
+#[repr(C)]
+#[derive(Clone, Copy)]
```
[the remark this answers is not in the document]
ice1000: Seems that some codes are depending on the deriving of `Copy` and `Clone`.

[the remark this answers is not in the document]
BusyJay: For example, what does it mean when calling `Clone` on `GrpcSlice`? Does the refcount need to be changed?

src/call/mod.rs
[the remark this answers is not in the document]
BusyJay: I think the compiler is smart enough to optimize `.unwrap()`. Besides, it is confusing that there are both `read` and `read_directly` but they are just basically the same.

src/call/mod.rs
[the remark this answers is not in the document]
BusyJay: Difficulty is not deal breaker, correctness is. 🤣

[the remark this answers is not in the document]
ice1000: @BusyJay I mean, #225 includes everything in this PR. This PR is just a subset of #225 and I've added some useful `*_mut` getters so we can simply change some function calls to fix this.

If I have to fix this in this PR, I'll have to add some functions that are already present in #225 and I have extra merges to do in #225 .

But I'm fixing this anyway.

src/call/mod.rs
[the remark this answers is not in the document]
ice1000: `.unwrap()` means you're ignoring some errors. it's quite common practice to redirect error handling using `?`s or `map`s or `flat_map`s but if there's no error we'll never need to do those.

You're actually handling the never-present errors in #134 like
https://github.com/pingcap/grpc-rs/blob/8fff94905ca9bccc1875f7f5a9cae7d12aa17196/src/call/mod.rs#L149-L153
https://github.com/pingcap/grpc-rs/blob/8fff94905ca9bccc1875f7f5a9cae7d12aa17196/src/call/mod.rs#L122

These are all unnecessary. I believe these codes will get optimized by the compiler, but it's not readable comparing to this version:
https://github.com/pingcap/grpc-rs/blob/7372c787b1c5e98a9c34c9fcf3b8d4fd3a772b11/src/call/mod.rs#L164-L167

grpc-sys/src/lib.rs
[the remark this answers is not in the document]
ice1000: No, this is just a reference to a C object. Cloning a reference is quite common.

This struct is just for forwarding some C stuffs, AFAIK (if I understand your code correctly).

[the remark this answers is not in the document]
ice1000: > This struct is just for forwarding some C stuffs

This means you never create `GrpcSlice` yourself, it's created on C-side. You never destruct it yourself, it's destructed on C-side. Your snippet has initialized a `GrpcSlice` on Rust-side, it's absolutely and obviously unsafe. But we've never did that, do we?

This is why `grpc_slice_unref` is [not used](https://github.com/pingcap/grpc-rs/search?q=grpc_slice_unref&unscoped_q=grpc_slice_unref) on Rust-side.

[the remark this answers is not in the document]
ice1000: Sorry I've just read the pr again and I found that it's used [here](https://github.com/pingcap/grpc-rs/pull/222/files#diff-8c8308f2c854f39735f331e70f94d8d2R184). But that looks safe to me anyway

src/call/mod.rs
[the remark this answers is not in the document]
BusyJay: Those codes are just general implementation of read_to_end, feel free to use `unwrap` to replace them.

If `Read` trait is not implemented, then I'm OK to remove the `Result` wrapper. However, since `Read` trait is implemented, I prefer to just keep one unify interface.

grpc-sys/src/lib.rs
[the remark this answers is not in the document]
BusyJay: Please don't just focus on how grpcio uses the API. grpcio-sys is a public crate, people is free to depend on the ABI and implements what they want. Hence `Clone` and `Copy` need to be considered carefully here.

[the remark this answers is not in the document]
ice1000: @BusyJay This is a separated task, I'm not doing this in this PR.

src/call/mod.rs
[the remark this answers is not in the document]
ice1000: Ok, ok, ok.

src/call/mod.rs
```diff
@@ -146,23 +258,28 @@ impl BatchContext {
...
-            );
-            buffer.set_len(len);
+            buf = grpc_sys::grpcwrap_batch_context_take_recv_message(self.ctx);
+            if buf.is_null() {
+                return None;
+            }
+
+            reader = mem::zeroed();
+            assert_eq!(grpc_sys::grpc_byte_buffer_reader_init(&mut reader, buf), 1);
+            length = grpc_sys::grpc_byte_buffer_length(reader.buffer_out);
```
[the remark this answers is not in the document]
ice1000: He said grpc-sys is also a public library, so we may not consider it an internal stuff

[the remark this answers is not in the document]
BusyJay: > a public library

What I'm talking about is whether GrpcSlice should derive `Clone` and `Copy`.

> Why not wrap it in C

Then you need a pointer to access it, which means allocation.

grpc-sys/src/lib.rs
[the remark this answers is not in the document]
BusyJay: Why is it a separated task? Is `Clone` and `Copy` necessary? What would happen if just removing them?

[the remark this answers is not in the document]
ice1000: You'll get compilation error due o implicit copy is no longer possible

ice1000:
Wowwww!!! Twavis CI passed, sooo lowely UwU

overvenus:
PTAL @hicqu, thanks!

siddontang:
PTAL @BusyJay

siddontang:
can benchmark result?

siddontang:
please do the benchmark and record in confluence. 
/cc @overvenus

overvenus:
Benchmarked in streaming_qps_unconstrained and unary_qps_unconstrained.


QPS: zcp recv means zero receive
screenshot_20180929_175000

Latency: zcp recv means zero receive
screenshot_20180929_175019

Zcp is faster than master with epllsig in rust_protobuf_async_streaming_qps_unconstrained.

Also, epollex is faster than epollsig in gRPC 1.42.

Logs: 

[zcprecv.epollex.log](https://github.com/pingcap/grpc-rs/files/2430501/zcprecv.epollex.log)
[zcprecv.epollsig.log](https://github.com/pingcap/grpc-rs/files/2430502/zcprecv.epollsig.log)
[masterrecv.epollex.log](https://github.com/pingcap/grpc-rs/files/2430503/masterrecv.epollex.log)
[masterrecv.epollsig.log](https://github.com/pingcap/grpc-rs/files/2430504/masterrecv.epollsig.log)

src/call/mod.rs
```diff
@@ -108,6 +110,116 @@ impl RpcStatus {
...
+                return Ok(&[]);
+            }
+            let mut len = 0;
+            unsafe {
+                match self.slice {
+                    None => self.slice = Some(mem::zeroed()),
+                    Some(s) => grpc_sys::grpc_slice_unref(s),
+                }
+                let slice = self.slice.as_mut().unwrap();
+                let code = grpc_sys::grpc_byte_buffer_reader_next(&mut self.reader, slice);
```
hicqu: If we get nothing in `grpc_byte_buffer_reader_next`, `slice` will be tought or not?

src/call/mod.rs
hicqu: It's not a problem. The copy happens when other components use the Reader.

src/call/mod.rs
ice1000:
```cpp
/** Updates \a slice with the next piece of data from from \a reader and returns
 * 1. Returns 0 at the end of the stream. Caller is responsible for calling
 * grpc_slice_unref on the result. */
GRPCAPI int grpc_byte_buffer_reader_next(grpc_byte_buffer_reader* reader,
                                         grpc_slice* slice);
```

If there's nothing to receive, we're already returned. So if we get nothing here, we should do a bug report to grpc.

ice1000:
@kennytm @hicqu @overvenus PTAL

siddontang:
seem zero has no any big difference in unary_qps_unconstrained? @overvenus

src/call/mod.rs
```diff
@@ -108,6 +110,116 @@ impl RpcStatus {
...
+/// To achieve zero-copy, use the BufRead API `fill_buf` and `consume`
+/// to operate the reader.
+pub struct MessageReader {
+    buf: *mut GrpcByteBuffer,
+    reader: GrpcByteBufferReader,
+    slice: Option<GrpcSlice>,
+    // Hack: apparently its lifetime depends on `self.slice`. However
+    // it's not going to be accessed by others directly, hence it's safe
+    // to mark it static here.
+    bytes: &'static [u8],
```
kennytm: Will `MessageReader` ever be moved? If `bytes` refer to an inlined GrpcSlice this would point to an invalid location?

ice1000: Reasonable. I'll take a look at it soon but friendly ping @BusyJay here

ice1000: In `fill_buf` (zero copy read), the `self.bytes` is returned a borrowed value. Once the `MessageReader` gets moved, you will no longer ba able to access this returned value, so it becomes invalid *and inaccessible*. By the way it's expected to call `consume` each time before `fill_buf`, it's rather safe in actual usage. Otherwise I expect a segfault (how can we check for that circumstance?).

ice1000: @kennytm Ping

kennytm: @ice1000 At the minimum please add the explanation in the code. 

----

Also, this should be possible without the unsafe hack, by eliminating the self-reference:

* you could separate `MessageReader` into two types:

    ```rust
    struct MessageBuffer { 
        buf: *mut GrpcByteBuffer,
        slice: Option<GrpcSlice>,
    }

    struct MessageReader<'b> {   // <-- note lifetime
        buf: &'b mut MessageBuffer,
        bytes: &'b [u8],
        length: usize,
    }
    ```

    however this probably involves a lot of plumbing to push through this extra lifetime

* alternatively, you could record the *offset* instead

    ```rust
    struct MessageReader {
        buf: *mut GrpcByteBuffer,
        slice: Option<GrpcSlice>,
        offset: usize,
        capacity: usize,  
        length: usize,
    }
    ```

    and change all references to `self.bytes` into `slice::from_raw_parts(GRPC_SLICE_START_PTR(slice) + offset, capacity - offset)`. This might cause a little overhead.

The point is that, by removing these lifetime-related unsafe code, we don't need to manually verify when the MessageReader becomes inaccessible or needs a human to check the `consume` → `fill_buf` order (which is error prone), the compiler itself can assure us these wrong usages will not cause big troubles (panic instead of segfault in the worst case).

ice1000: > At the minimum please add the explanation in the code.

@BusyJay has already explained (but I'll add more detailed comments later): https://github.com/pingcap/grpc-rs/blob/e012efae41a31b17823429d4ec6419cf45e85920/src/call/mod.rs#L121-L123

> you could separate MessageReader into two types

Are you sure? If so, how can we return a `io::Result<&[u8]>` in `fill_buf` (since there's no lifetime parameter can be specified in the trait)?

> alternatively, you could record the offset instead

Then, if an inlined buffer is moved (as you said, the pointer becomes invalid), will `slice::from_raw_parts(GRPC_SLICE_START_PTR(slice) + offset, capacity - offset)` return a valid pointer? BTW we're not using offset here.

The point is that, we don't usually call `consume`, `fill_buf` directly. Instead, we use wrappers provided by the `BufRead` trait, which are correctly implemented (at least in the codec mod)

kennytm: @ice1000 It only explains the `'static` lifetime 😛. What I mean specifically is to explain that moving MessageReader will make it inaccessible.

<details><summary>More reply, not part of the LGTM blocker</summary>

> Are you sure? If so, how can we return a io::Result<&[u8]> in fill_buf (since there's no lifetime parameter can be specified in the trait)?

`'a` outlives `'self` so you could just return `self.bytes`.

```rust
type MessageBuffer = [u8; 128]; // for demo only
struct MessageReader<'b> {
    buf: &'b mut MessageBuffer,
    bytes: &'b [u8],
    length: usize,
}
impl<'b> MessageReader<'b> {
    fn fill_buf(&mut self) -> std::io::Result<&[u8]> {
        Ok(self.bytes)
    }
}
```

> Then, if an inlined buffer is moved (as you said, the pointer becomes invalid), will slice::from_raw_parts(GRPC_SLICE_START_PTR(slice) + offset, capacity - offset) return a valid pointer? 

You *recompute* `GRPC_SLICE_START_PTR(slice)` every time you need to use the bytes, and thus you'll always get the correct address no matter where you're going to move the MessageReader.

> The point is that, we don't usually call consume, fill_buf directly. Instead, we use wrappers provided by the BufRead trait, which are correctly implemented (at least in the codec mod)

`MessageReader` is a public struct and `BufRead` is a public trait so *any third-party user* using grpc-rs can call `consume()` and `fill_buf()` in any order they like. Anyway it's not a big issue.

</details>

src/call/mod.rs
hicqu: Got it.

grpc-sys/src/lib.rs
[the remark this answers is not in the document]
hicqu: I think the conversation should be not resolved. I think we need to remove `Clone` and `Copy` derive at least. Any updates?

src/call/mod.rs
hicqu: Any updates?

ice1000: @hicqu I was busy with fixing formatting issue last night, will do it tonight

ice1000: Done

ice1000: Already fixed, could you please take a look? @kennytm

src/call/mod.rs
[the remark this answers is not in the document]
ice1000: Fine

src/call/mod.rs
hicqu: In the fix what you clone is a `&'static [u8]`. I think the problem is not resolved. I think the correct solution is
```
+        self.bytes = self.slice.as_mut().map_or(&[], |s| unsafe {
+            let mut len = 0;
+            let ptr = grpc_sys::grpcwrap_slice_raw(s, &mut len);
+            slice::from_raw_parts(ptr as _, len)
+        });
```
What you need is just fix the panic.

ice1000: Fine. There's too much checks to do if I remove the `'static` (and add another field).

ice1000: I'll do it tomorrow.

kennytm: What I meant was entirely remove the `bytes` field 🙂

ice1000: @kennytm No, you can't. `consume`.

kennytm: @ice1000 That's why I said you record the *offset* in https://github.com/pingcap/grpc-rs/pull/222#discussion_r224313247. `consume` will bump the offset.

ice1000: How reasonable

grpc-sys/src/lib.rs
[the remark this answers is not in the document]
ice1000: @BusyJay @hicqu Currently the impl of `MessageReader` is C-style, its maintainability is bad but its logic is correct. I'll refactor this part in a separate PR to avoid taking too much time by further code review&discussion.

BTW PTAL

ice1000:
Fixed. BTW `cargo fmt` is not displaying anything locally.

hicqu:
LGTM.

ice1000:
I've finally fixed all the formatting issues from https://travis-ci.org/pingcap/grpc-rs/jobs/443011498

ice1000:
Oh no! 
image

:sob:

PR #231, opened 2018-09-25
WIP: Adding Server::socket_mutator method

dwerner:
Currently, there is no way to utilize the grpc_socket_mutator struct+vtable from rust. This PR adds support for doing that from ServerBuilder, and in the process I ended up separating out ChannelBuilder's members that pertained to building ChannelArgs into their own builder. There's also some C++ code in grpc_wrap.cc that needed to be added to hook a callback in.

This PR provides an api where the user, upon using ServerBuilder, can simply pass a closure which is called with the fd from the grpc_socket_mutator vtable.

Please let me know thoughts on reworking this into something acceptable for merging. Thanks for the great library!

### DIFF
--- a/grpc-sys/grpc_wrap.cc
+++ b/grpc-sys/grpc_wrap.cc
@@ -57,6 +57,9 @@
 #include <grpc/support/string_util.h>
 #include <grpc/support/thd_id.h>
 
+#include "src/core/lib/gpr/useful.h"
+#include "src/core/lib/iomgr/socket_mutator.h"
+
 #ifdef GRPC_SYS_SECURE
 #include <grpc/grpc_security.h>
 #endif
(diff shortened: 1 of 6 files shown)


BusyJay:
Interesting. I'm not familiar with `grpc_socket_mutator`, can you provide more information about it? Like what's the situation that it can be useful for?

dwerner:
In the grpc core library, `grpc_socket_mutator` is essentially a singleton hook that allows someone to mutate a file-descriptor as they are opened by the underlying `iomgr`. This then allows someone using the library to call `setsockopt()` or similar on a file descriptor. This could be considered a bit edge-case, but sometimes building a high-performance, low-latency server requires hooks to low levels like this. The only documentation I can find on it are the unit tests in grpc++ and some doxygen ( https://grpc.io/grpc/cpp/classgrpc_1_1_channel_arguments.html#a520690f499be85159a20200ddb986a96 )

There are other similar APIs (`grpc_socket_factory`, for instance) that I have not yet explored, but follow a similar model of ref-counted struct holding a vtable. These are passed in as channel arguments as `GRPC_ARG_POINTER`, similar to this example for `grpc_socket_mutator`.

For my use case, I need to be able to call `libc::setsockopt()` with `TCP_FASTOPEN` and `TCP_DEFER_ACCEPT`. Or, for instance, disable `SO_REUSEPORT` in debugging.

Some obvious deficiencies that I plan to address are: cleanup - I leak the `Box<Fn>` that C holds as a `void*` currently, and that would need to hook into `Server::drop`.

grpc-sys/grpc_wrap.cc
```diff
@@ -57,6 +57,9 @@
 #include <grpc/support/string_util.h>
 #include <grpc/support/thd_id.h>
 
+#include "src/core/lib/gpr/useful.h"
+#include "src/core/lib/iomgr/socket_mutator.h"
```
BusyJay: This will prevent users to link to local grpc library. Only public headers can be depended on.

dwerner: Ah, thanks, this should be accessible through `grpc-sys/grpc/include/grpc/impl/codegen/grpc_types.h` instead.

dwerner: On a more detailed review, `grpc_socket_mutator` is a forward decl in grpc_types.h, however `grpc_socket_mutator_vtable` is not, and so part of the required api is not public.

dwerner:
Because the required apis in grpc aren't (completely) public (namely `grpc_socket_mutator_vtable`) I'm closing this PR.

PR #235, opened 2018-09-30
[Do not merge] Try reserving sized data in advance

ice1000:
Signed-off-by: ice1000 <ice1000kotlin@foxmail.com>

This is just a test for trying to reserve a computed size of capacity for the `Vec`.
This can reduce possible unnecessary allocations and copies.

Created PR so we can discuss the benchmark, tests, etc.

### DIFF
--- a/src/call/client.rs
+++ b/src/call/client.rs
@@ -113,7 +113,7 @@ impl Call {
         mut opt: CallOption,
     ) -> Result<ClientUnaryReceiver<Resp>> {
         let call = channel.create_call(method, &opt)?;
-        let mut payload = vec![];
+        let mut payload = Vec::new();
         (method.req_ser())(req, &mut payload);
         let cq_f = check_run(BatchType::CheckRead, |ctx, tag| unsafe {
             grpc_sys::grpcwrap_call_start_unary(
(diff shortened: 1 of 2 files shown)


src/call/client.rs
```diff
@@ -167,7 +167,7 @@ impl Call {
         mut opt: CallOption,
     ) -> Result<ClientSStreamReceiver<Resp>> {
         let call = channel.create_call(method, &opt)?;
-        let mut payload = vec![];
+        let mut payload = Vec::new();
```
ice1000: Allocating 0-sized memory because we'll reserve a size later and we don't need the first allocation.

Ref: https://stackoverflow.com/q/2022335/7083401

src/codec.rs
```diff
@@ -36,13 +36,16 @@ pub struct Marshaller<T> {
...
 pub mod pb_codec {
-    use protobuf::{self, Message};
+    use protobuf::{self, Message, stream::WithCodedOutputStream};
 
     use error::Result;
 
     #[inline]
     pub fn ser<T: Message>(t: &T, buf: &mut Vec<u8>) {
-        t.write_to_vec(buf).unwrap()
+        t.check_initialized().unwrap();
```
ice1000: Inlined the implementation of `write_to_vec` and did a little modification.

ice1000:
@overvenus Can you do a benchmark?

ice1000:
Opened #238 , closing

PR #236, opened 2018-10-02
Allow '.' characters in metadata keys

heavypackets:
In the current state, grpc-rs does not allow '.' as a valid character in the key validation function, even though it is a valid character for keys in many other GRPC clients, notably Go, Ruby, and Java.

The most restrictive language I have seen regarding keys is within the official GPRC clients, this for [Java](https://grpc.io/grpc-java/javadoc/io/grpc/Metadata.Key.html) and [Go](https://godoc.org/google.golang.org/grpc/metadata#Pairs) -- it is in alignment to grpc-rs but also allows for '.' characters.

This is currently breaking the addition of a required security header (workload.spiffe.io=true) for the [SPIFFE standard](https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE_Workload_Endpoint.md#3-transport) workload API with this client.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -31,7 +31,12 @@ fn normalize_key(key: &str, binary: bool) -> Result<Cow<str>> {
         if b >= b'A' && b <= b'Z' {
             is_upper_case = true;
             continue;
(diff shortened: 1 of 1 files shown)


heavypackets:
Rolled back rebase of the current master as the build is currently broken.

PR #239, opened 2018-10-09
Optionally use openssl instead of boring ssl.

bacek:
This prevent link/runtime failure when you have a crate which uses both
grpc-rs and openssl-rs.

Based on feedback from #186. Build against fresh master.

### DIFF
--- a/grpc-sys/Cargo.toml
+++ b/grpc-sys/Cargo.toml
@@ -47,10 +47,12 @@ exclude = [
 
 [dependencies]
 libc = "0.2"
+# openssl = { version = "*", optional = true }
 
 [features]
 default = []
 secure = []
+openssl = ["secure"]
 
 [build-dependencies]
 cc = "1.0"
--- a/grpc-sys/Cargo.toml
+++ b/grpc-sys/Cargo.toml
@@ -47,7 +47,6 @@ exclude = [
 
 [dependencies]
 libc = "0.2"
-# openssl = { version = "*", optional = true }
 
 [features]
 default = []
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,3 +48,5 @@ script:
   - cargo build
   - cargo test --all
   - GRPCIO_SYS_USE_PKG_CONFIG=1 cargo test --all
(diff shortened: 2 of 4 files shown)


grpc-sys/Cargo.toml
```diff
@@ -47,10 +47,12 @@ exclude = [
 
 [dependencies]
 libc = "0.2"
+# openssl = { version = "*", optional = true }
```
overvenus: It can be removed.

overvenus:
Cmake can not find openssl on macOS. https://travis-ci.org/pingcap/grpc-rs/jobs/439426196#L1504

```
--- stderr
CMake Error at /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.11.4/share/cmake/Modules/FindOpenSSL.cmake:390 (find_package_handle_standard_args)
  cmake/ssl.cmake:38 (find_package)
  CMakeLists.txt:119 (include)
```

grpc-sys/Cargo.toml
bacek: Ooops. Removed.

bacek:
OSX build are fixed now. rust-beta are failing independently from openssl.

bacek:
@overvenus Can you merge it? Looks like automatic merge is blocked by rust-beta Travis failure.

overvenus:
I can rerun travis CI, but still need one approval from maintainer in order to merge.

PTAL @BusyJay

bacek:
Is RemoteStopped legitimate response in case of dropping tx/rx channel? Because looks like there is a race condition between Sink::start_send and async::check_alive.

Issue #242, opened 2018-10-19
Unstable test


ice1000:
Currently the unit test is not so stable.
This code sometimes get crashed:

https://github.com/pingcap/grpc-rs/blob/c93d7d4b9b18b66187a5efd6c7d7eec8d38f192f/tests/cases/cancel.rs#L126-L128

nrc:
E.g.:

```
thread 'cases::cancel::test_client_cancel_on_dropping' panicked at 'expected cancel, but got: RemoteStopped', tests/cases/cancel.rs:127:24
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:339
   9: tests::cases::cancel::check_cancel
             at tests/cases/cancel.rs:127
  10: tests::cases::cancel::test_client_cancel_on_dropping
             at tests/cases/cancel.rs:185
  11: tests::cases::cancel::test_client_cancel_on_dropping::{{closure}}
             at tests/cases/cancel.rs:150
  12: core::ops::function::FnOnce::call_once
             at /rustc/bc2e84ca0939b73fcf1768209044432f6a15c2e5/src/libcore/ops/function.rs:231
  13: <alloc::boxed::Box<F> as core::ops::function::FnOnce<A>>::call_once
             at /rustc/bc2e84ca0939b73fcf1768209044432f6a15c2e5/src/liballoc/boxed.rs:766
  14: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:82
  15: std::panicking::try
[...]

PR #238, opened 2018-10-09
Zero copy for send

ice1000:
## TODOs

+ [X] Fix benchmark compilation
+ [x] Fix link error (multiple symbol definitions)
+ [x] Fix `size_t` 6 in Rust becomes `size_t` in C 140184721559552 via FFI (140184721559552+6 is 2^47)
+ [x] Sync master changes
+ [x] Improve `GrpcSlice`
+ [x] Improve `GrpcByteBuffer`
+ [x] (New!) Support bytes in prost

## Now (updated at: 23, April 2019)

Since @nrc has introduced prost support (instead of rust-protobuf) which is natively using "bytes", we no longer have that one copy for send which is stated to be necessary below.
Yay!

## Before:

+ User code deserialize (particularly, the protobuf library), producing a list of `&[u8]`, we copied the data into a `Vec<u8>`, this is done internally in the protobuf library (**copy 0**)
+ We create a `grpc_slice` (a ref-counted single string) by copying the `Vec<u8>` mentioned above (**copy 1**)
+ We create a `grpc_byte_buffer` (a ref-counted list of strings) from only one `grpc_slice`
+ Send

**Total copy: 2**

## After:

+ User code deserialize (particularly, the protobuf library), producing a list of `&[u8]`, we copied each produced `&[u8]` and collect them into a `grpc_byte_buffer` (**copy 0**)
+ Send

**Total copy: 1**

The reason why still one copy:

+ Rust manages its own allocated memory'a lifetime
+ C-side uses a ref-count mechanism to manage the lifetime of objects
+ We either need to longer the Rust objects' lifetime or copy once
+ Tried to longer Rust objects' lifetime in #225 
  + Result: grpc supports buffer-hint which requires a much longer lifetime, it's too hard to manage the lifetime in such case
  + This is why I closed #225

PR #246, opened 2018-10-22
server: check fail eagerly

### LINKED ISSUE #175: Is there a way to detect a closed connection between sending stuff to the sink
If I try to send stuff in the sink and the connection is closed I'll get an error but I'm wondering if there's a way to detect the connection even if I don't send stuff into the sink for a long time.

I want to use this crate for a long running pub sub. In some case, a lot of time (hours) can occurs between each messages.

In this exemple I send something every 60 secs. If the client closes the connection, I only have an error the next time the 60 secs timer fires and I would like to know right away if possible.

```rust
struct TestService;

impl TestCase for TestService {
    fn sub(&self, ctx: RpcContext, _req: Empty, sink: ServerStreamingSink<Notification>) {
        println!("new connection");

        let (tx, rx) = futures::sync::mpsc::unbounded();

        thread::spawn(move || loop {
            println!("send");
            tx.unbounded_send(1).unwrap();
            thread::sleep(time::Duration::from_secs(60));
        });

        let f = rx.map(|v| {
            let mut n = Notification::new();
            n.set_value(v.to_string());
            (n, WriteFlags::default())
        }).map_err(|()| err_msg("error rx"))
            .forward(sink)
            .map(|_| ())
            .map_err(|e| {
                println!("error: {}", e);
                ()
            });

        ctx.spawn(f);
    }
}
```

BusyJay:
It's possible that server stream is flushing messages when the connection is closed. So we need to check if the call is still valid inside `poll_complete`.

Close #175.

### DIFF
--- a/tests/cases/cancel.rs
+++ b/tests/cases/cancel.rs
@@ -11,13 +11,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::mpsc as std_mpsc;
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::Duration;
-use std::sync::mpsc as std_mpsc;
 
-use futures::{future, stream as streams, Async, Future, Poll, Sink, Stream};
 use futures::sync::mpsc;
+use futures::{future, stream as streams, Async, Future, Poll, Sink, Stream};
 use grpcio::*;
 use grpcio_proto::example::route_guide::*;
 use grpcio_proto::example::route_guide_grpc::*;
@@ -68,11 +68,11 @@ impl RouteGuide for CancelService {
             thread::sleep(Duration::from_secs(5));
         });
 
-        let f = rx.map(|_| {
-            let f = Feature::new();
-            (f, WriteFlags::default())
-        })
-            .forward(sink.sink_map_err(|_| ()))
+        let f = rx
+            .map(|_| {
+                let f = Feature::new();
+                (f, WriteFlags::default())
+            }).forward(sink.sink_map_err(|_| ()))
             .map(|_| ())
             .map_err(|_| ())
             .then(move |_| {
@@ -265,7 +265,7 @@ fn test_early_exit() {
         let l = client.list_features(&rect).unwrap();
         let f = l.into_future();
         match f.wait() {
-            Ok((Some(_), _)) => {},
+            Ok((Some(_), _)) => {}
             Ok((None, _)) => panic!("should have result"),
             Err((e, _)) => panic!("unexpected error {:?}", e),
         };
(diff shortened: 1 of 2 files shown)


kennytm:
CI is failing this error looks legit.

```
---- cases::cancel::test_client_cancel_on_dropping stdout ----
thread 'cases::cancel::test_client_cancel_on_dropping' panicked at 'expected cancel, but got: RemoteStopped', tests/cases/cancel.rs:121:24

   7: tests::cases::cancel::check_cancel
             at tests/cases/cancel.rs:121
   8: tests::cases::cancel::test_client_cancel_on_dropping
             at tests/cases/cancel.rs:160
   9: tests::cases::cancel::test_client_cancel_on_dropping::{{closure}}
             at tests/cases/cancel.rs:142
```

ice1000:
The error is due to #242

tests/cases/cancel.rs
```diff
@@ -27,13 +31,15 @@ type RouteChatHandler =
 
 #[derive(Clone)]
 struct CancelService {
+    list_feature_listener: Arc<Mutex<Option<std_mpsc::Sender<()>>>>,
```
overvenus: What about `Handler<Sender<()>>`?

tests/cases/cancel.rs
```diff
@@ -222,3 +254,21 @@ fn test_server_cancel_on_dropping() {
...
     check_cancel(rx);
 }
+
+#[test]
+fn test_early_exit() {
+    let (service, client, _server) = prepare_suite();
+    let (tx, rx) = std_mpsc::channel();
+    *service.list_feature_listener.lock().unwrap() = Some(tx);
+    let rect = Rectangle::new();
+    {
```
overvenus: Could you remove `{`?

tests/cases/cancel.rs
BusyJay: It's just a `Sender` rather than a `Box`.

BusyJay:
PTAL

Hoverbear:
@ice1000 did this fix #242?

ice1000:
@Hoverbear Still happens to me

```
failures:

---- cases::cancel::test_client_cancel_on_dropping stdout ----
thread 'cases::cancel::test_client_cancel_on_dropping' panicked at 'expected cancel, but got: RemoteStopped', tests/cases/cancel.rs:121:24
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    cases::cancel::test_client_cancel_on_dropping

test result: FAILED. 7 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

```

(With the branch in #238 merged into master)

PR #251, opened 2018-10-25
Make the generated *Client classes clonable; This requires making grpcio::Client clonable too

thomasetter:
This is especially useful for services connecting to other services as it allows to just make the client a member of the struct which implements the *Service trait while using `#[derive(Clone)]`.

Tests:
`cargo test` at the root folder and built one of my own projects with this. Do you have some other tests you would like me to run?


ice1000:
No, `cargo test` is fine. Will look into this.

ice1000:
@overvenus @BusyJay Please take a look to see if there's anything improper.

PR #252, opened 2018-10-29
Created a wrapper for GrpcSlice and GrpcByteBuffer

ice1000:


grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,184 @@ pub union GrpcSliceData {
...
+}
+
+/// A light wrapper with constructor, destructor for
+/// [GrpcSliceRaw]
+pub struct GrpcSlice {
+    raw: GrpcSliceRaw,
+}
+
+impl GrpcSlice {
+    pub fn new(len: usize) -> Self {
```
overvenus: How about `with_capacity`? Like `Vec::with_capacity`?

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,184 @@ pub union GrpcSliceData {
...
+
+    pub unsafe fn as_ptr(&self) -> GrpcSliceRaw {
+        self.raw.clone()
+    }
+}
+
+/// Increase the ref count of the slice when cloning
+impl Clone for GrpcSlice {
+    fn clone(&self) -> Self {
+        unsafe {
```
overvenus: Looks like we can reuse the returned `GrpcSliceRaw`.

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,184 @@ pub union GrpcSliceData {
...
+impl GrpcSlice {
+    pub fn new(len: usize) -> Self {
+        unsafe { From::from(grpc_slice_malloc(len)) }
+    }
+
+    pub fn len(&self) -> usize {
+        unsafe { grpcwrap_slice_length(&self.raw) }
+    }
+
+    pub fn as_slice(&self, offset: usize) -> &[u8] {
```
overvenus: It does not intuition `as_slice` requires a offset. Could you rename it?

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,184 @@ pub union GrpcSliceData {
...
+        unsafe {
+            let mut len = 0;
+            let ptr = grpcwrap_slice_raw_offset(&self.raw, offset, &mut len);
+            slice::from_raw_parts(ptr as _, len)
+        }
+    }
+
+    pub unsafe fn as_ptr(&self) -> GrpcSliceRaw {
```
overvenus: `GrpcSliceRaw` is not a ptr.

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,184 @@ pub union GrpcSliceData {
...
+/// Trivial copy. We need to copy the inlined-array, which means we can't
+/// Simply `#[derive(Clone)]`.
+/// Here we use the API provided by gRPC.
+impl Clone for GrpcSliceRaw {
+    fn clone(&self) -> Self {
+        unsafe { grpcwrap_slice_copy(self as _) }
+    }
+}
+
+impl<'a> From<&'a [i8]> for GrpcSliceRaw {
```
kennytm: Why is this impl added 🤔.

src/call/mod.rs
```diff
@@ -764,27 +715,23 @@ impl SinkBase {
...
-    unsafe fn make_message_reader(source: &[u8], nslice: usize) -> MessageReader {
-        let mut slices = Vec::with_capacity(nslice);
-        for _ in 0..nslice {
-            let len = source.len();
-            let mut slice = grpc_sys::grpc_slice_from_copied_buffer(source.as_ptr() as _, len);
-            slices.push(slice);
+    fn make_message_reader(source: &[u8], n_slice: usize) -> MessageReader {
+        let mut slices = Vec::with_capacity(n_slice);
+        for _ in 0..n_slice {
+            slices.push(From::from(source));
```
kennytm: why can't this simply be 

```rust
let mut slices = vec![source.into(); n_slice];
```

grpc-sys/src/lib.rs
ice1000: Do you have some suggestions about the name?

grpc-sys/src/lib.rs
ice1000: Dunno. Will delete it later

src/call/mod.rs
ice1000: I don't know there's such API, will refactor later.

grpc-sys/src/lib.rs
overvenus: So in order to clone `GrpcSlice`, we have to clone `GrpcSliceRaw` twice?

grpc-sys/src/lib.rs
overvenus: Ideally we should impl `Index<Range<usize>>` for `Index<RangeFrom<usize>>`, but for simplicity I prefer `range_from`.

https://doc.rust-lang.org/std/ops/trait.Index.html

grpc-sys/src/lib.rs
overvenus: How about `into_raw`?

https://doc.rust-lang.org/std/index.html?search=into_raw

kennytm: `as_raw` since this doesn't take ownership and is a cheap conversion.

ice1000: I've renamed it locally already, just forgot to push it.

ice1000: Renamed into `as_raw`

grpc-sys/src/lib.rs
ice1000: `GrpcSliceRaw` is of value semantics. Clone first to inc the ref count, clone second time to create a new ref.

To reduce this (although very cheap) copy, we can make a `grpcwrap_slice_ref` which takes a pointer instead of a value.

BusyJay:
I suggest to merge xxxRaw with xxx.

grpc-sys/src/lib.rs
ice1000: Can't delete, required in the test

kennytm: That's odd, which test produces an `&[i8]` (or `&[c_char]`)? This normally shouldn't happen in Rust

siddontang:
please add a description for your PR at first :-)

ice1000:
> I suggest to merge xxxRaw with xxx.

This is my original intent, but:
+ In some cases we still need the `*Raw` version
+ We need the raw copy of `*Raw`s to implement `Clone` with ref-count

grpc-sys/src/lib.rs
ice1000: Sorry, deleting it works now.

ice1000:
> please add a description for your PR at first :-)

It's a mistake, my keyboard was choking when I opened this PR. Added description as a comment because I consider editing the description improper (especially when I'm able to add a comment whose location is very close to the title):

image

grpc-sys/src/lib.rs
ice1000: Re-used.

ice1000:
@overvenus @kennytm @BusyJay @siddontang I'm all set.

ice1000:
Writing down some notes:

```

error: item `GrpcSlice` has a public `len` method but no corresponding `is_empty` method
   --> grpc-sys/src/lib.rs:408:1
    |
408 | / impl GrpcSlice {
409 | |     pub fn with_capacity(len: usize) -> Self {
410 | |         unsafe { From::from(grpc_slice_malloc(len)) }
411 | |     }
...   |
427 | |     }
428 | | }
    | |_^
    |
    = note: `-D clippy::len-without-is-empty` implied by `-D warnings`
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#len_without_is_empty
error: item `GrpcByteBufferReader` has a public `len` method but no corresponding `is_empty` method
   --> grpc-sys/src/lib.rs:481:1
    |
481 | / impl GrpcByteBufferReader {
482 | |     pub fn len(&self) -> usize {
483 | |         unsafe { grpc_byte_buffer_length(self.buffer_out) }
484 | |     }
...   |
493 | |     }
494 | | }
    | |_^
    |
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#len_without_is_empty
error: defining a method called `next` on this type; consider implementing the `std::iter::Iterator` trait or choosing a less ambiguous name
   --> grpc-sys/src/lib.rs:486:5
    |
486 | /     pub fn next(&mut self) -> GrpcSlice {
487 | |         unsafe {
488 | |             let mut raw: GrpcSliceRaw = mem::zeroed();
489 | |             let code = grpc_byte_buffer_reader_next(self, &mut raw);
...   |
492 | |         }
493 | |     }
    | |_____^
    |
    = note: `-D clippy::should-implement-trait` implied by `-D warnings`
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#should_implement_trait
```

ice1000:
Removed `Raw` for `GrpcSlice`, applied suggestions from clippy.
PTAL @overvenus @BusyJay @kennytm

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,151 @@ pub union GrpcSliceData {
...
+        }
+    }
+}
+
+/// Increase the ref count of the slice when cloning
+impl Clone for GrpcSlice {
+    fn clone(&self) -> Self {
+        unsafe {
+            grpcwrap_slice_ref(self);
+            grpcwrap_slice_copy(self)
```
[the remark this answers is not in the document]
ice1000: return a copy of the reference. Why not?
How am I supposed to do the actual clone then?

grpc-sys/src/lib.rs
```diff
@@ -409,14 +536,23 @@ pub enum GrpcCallDetails {}
...
 pub const GRPC_MAX_COMPLETION_QUEUE_PLUCKERS: usize = 6;
 
 extern "C" {
+    pub fn grpcwrap_slice_copy(slice: *const GrpcSlice) -> GrpcSlice;
+    pub fn grpcwrap_slice_ref(slice: *const GrpcSlice) -> GrpcSlice;
+    pub fn grpc_empty_slice() -> GrpcSlice;
+    pub fn grpc_slice_malloc(len: usize) -> GrpcSlice;
+    pub fn grpc_slice_ref(slice: GrpcSlice) -> GrpcSlice;
+    pub fn grpc_slice_unref(slice: GrpcSlice);
+    pub fn grpcwrap_slice_unref(slice: *const GrpcSlice);
```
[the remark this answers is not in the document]
ice1000: Those are added separately, I organize them according to their functionality.

Make sense, will refactor

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,151 @@ pub union GrpcSliceData {
...
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+impl Iterator for GrpcByteBufferReader {
+    type Item = GrpcSlice;
+
+    fn next(&mut self) -> GrpcSlice {
```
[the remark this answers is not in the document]
ice1000: Ok

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,151 @@ pub union GrpcSliceData {
...
+        unsafe {
+            reader = mem::zeroed();
+            let init_result = grpc_byte_buffer_reader_init(&mut reader, src.raw);
+            assert_eq!(init_result, 1);
+        }
+        reader
+    }
+}
+
+pub struct GrpcByteBuffer {
```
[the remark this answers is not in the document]
ice1000: It's better to have an automatic destructed wrapper when we actually need one, although it's used only once (create one for future convenience)

grpc-sys/src/lib.rs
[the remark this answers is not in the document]
ice1000: Sorry.

overvenus:
Well done! LGTM, though clippy complains https://travis-ci.org/pingcap/grpc-rs/jobs/450134527#L811

ice1000:
Fixed

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,144 @@ pub union GrpcSliceData {
...
+    fn drop(&mut self) {
+        unsafe {
+            grpcwrap_slice_unref(self);
+        }
+    }
+}
+
+impl<'a> From<&'a [u8]> for GrpcSlice {
+    fn from(data: &'a [u8]) -> Self {
+        unsafe { grpc_slice_from_copied_buffer(data.as_ptr() as _, data.len()) }
```
hicqu: I think the buffer won't be copyed by the function, right?

hicqu: It will copy it. ignore it.

ice1000:
@hicqu @overvenus Any approving reviews? :wink:

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,144 @@ pub union GrpcSliceData {
...
+            let mut slice = Default::default();
+            let code = grpc_byte_buffer_reader_next(self, &mut slice);
+            debug_assert_ne!(code, 0);
+            slice
+        }
+    }
+}
+
+impl<'a> From<&'a mut GrpcByteBuffer> for GrpcByteBufferReader {
+    fn from(src: &'a mut GrpcByteBuffer) -> Self {
```
hicqu: Why not let `GrpcByteBufferReader` holds `GrpcByteBuffer` instead of some raw pointers? So we don't need `_buf` in the `MessageReader`. And so I think we can merge `GrpcByteBuffer` and `Raw`.

overvenus:
@BusyJay PTAL, in particular `GrpcByteBufferRaw` and `GrpcByteBuffer`.

grpc-sys/src/lib.rs
ice1000: No, we need `Raw` version. In some existing code, we need to hold it as a reference and we cannot destruct it.
Also, `GrpcByteBufferReader` is a struct from C side, it's probably a bad idea to change its structure.

ice1000: Also, we always operate `GrpcByteBuffer` as a pointer. If we want to RAII it, we need to create the complete struct in Rust, which will bring extra work (this struct depend on other structs, and recurse).

hicqu: So why we need `GrpcByteBuffer`? And if GrpcByteBuffer has only 1 pointer field, their memory layout are exactly equal.

src/call/mod.rs
```diff
@@ -274,6 +224,15 @@ impl BatchContext {
         self.ctx
     }
 
+    pub fn take_recv_message(&self) -> Option<GrpcByteBuffer> {
+        let ptr = unsafe { grpc_sys::grpcwrap_batch_context_take_recv_message(self.ctx) };
```
hicqu: Where will the pointer will be de-alloc?

ice1000: It's wrapped into a `GrpcByteBuffer`, and will be de-alloc by
```rust
impl Drop for GrpcByteBuffer {
    fn drop(&mut self) {
        unsafe { grpc_byte_buffer_destroy(self.raw) }
    }
}
```

ice1000:
> I suggest to leave it to users. Don't overdo it. And this is not consistent with the style of other types in this file.

Created an external wrapper in call/mod.rs

grpc-sys/src/lib.rs
```diff
@@ -379,24 +379,144 @@ pub union GrpcSliceData {
...
 #[repr(C)]
-#[derive(Clone, Copy)]
 pub struct GrpcSlice {
     ref_count: *mut GrpcSliceRefCount,
     data: GrpcSliceData,
 }
 
+impl GrpcSlice {
+    pub fn with_capacity(len: usize) -> Self {
+        unsafe { grpc_slice_malloc(len) }
```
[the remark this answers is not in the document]
ice1000: Yes.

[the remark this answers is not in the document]
ice1000:
```c
grpc_slice grpc_slice_malloc_large(size_t length) {
  grpc_slice slice;

  /* Memory layout used by the slice created here:

     +-----------+----------------------------------------------------------+
     | refcount  | bytes                                                    |
     +-----------+----------------------------------------------------------+

     refcount is a malloc_refcount
     bytes is an array of bytes of the requested length
     Both parts are placed in the same allocation returned from gpr_malloc */
  malloc_refcount* rc = static_cast<malloc_refcount*>(
      gpr_malloc(sizeof(malloc_refcount) + length));

  /* Initial refcount on rc is 1 - and it's up to the caller to release
     this reference. */
  gpr_ref_init(&rc->refs, 1);

  rc->base.vtable = &malloc_vtable;
  rc->base.sub_refcount = &rc->base;

  /* Build up the slice to be returned. */
  /* The slices refcount points back to the allocated block. */
  slice.refcount = &rc->base;
  /* The data bytes are placed immediately after the refcount struct */
  slice.data.refcounted.bytes = reinterpret_cast<uint8_t*>(rc + 1);
  /* And the length of the block is set to the requested length */
  slice.data.refcounted.length = length;
  return slice;
}
```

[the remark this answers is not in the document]
ice1000:
```cpp
grpc_slice grpc_slice_malloc(size_t length) {
  grpc_slice slice;

  if (length > sizeof(slice.data.inlined.bytes)) {
    return grpc_slice_malloc_large(length);
  } else {
    /* small slice: just inline the data */
    slice.refcount = nullptr;
    slice.data.inlined.length = static_cast<uint8_t>(length);
  }
  return slice;
}
```

hicqu:
LGTM.

PR #254, opened 2018-11-05
Improve  RPC client clonable.

hicqu:
I want to write code like this:
```rust
let stream1 = client.stream_rpc1(); // A new RPC interface.
client.spawn(
    stream1.send_all(...).then(|r| match r {
        Err(e) => {
            // The new RPC is not supported by server side, so fallback to an old RPC.
            let stream2 = client.stream_rpc2(); // need a copy of `Client` here.
            stream2.send_all(...)
        }
})
```
`Client` is already `Arc`ed. So I don't want to Arc it again.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ all-features = true
 grpcio-sys = { path = "grpc-sys", version = "0.4.0" }
 libc = "0.2"
 futures = "^0.1.15"
-protobuf = { version = "~2.0", optional = true }
+protobuf = { version = "~2.1", optional = true }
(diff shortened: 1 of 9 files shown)


ice1000:
Any special requirement for copying the client? Like do we need to call some functions to ref and unref?

overvenus:
Duplicate with #251 ?

hicqu:
Closed by #251

hicqu:
This PR can improve the old implemention in #251 . Clone the `Arc` inside the `Client` is better than make a new call.

Issue #272, opened 2019-01-09
publish grpcio-compiler 0.4.2 to crates.io


dwerner:
I landed a [PR](https://github.com/pingcap/grpc-rs/pull/248) which brings in protobuf 2.1 serde_derive support, but I don't think a new version of grpcio-compiler has been published which exposes this feature.

overvenus:
@dwerner Sorry for the inconvenience. To support serde, we need to upgrade the protobuf to 2.1+ which is unfortunately not included in v0.4.2. As for grpcio-compiler, I dont think it need to upgrade, because it does not generate protobuf structs.

Will release 0.4.3 once we merge #273. Also, please let me know if #273 works for you.

dwerner:
#273 looks good to me! Thanks. I'll update the title to reflect 0.4.3 instead of 0.4.2

overvenus:
FYI, grpcio/compiler v0.4.3 is released.

PR #273, opened 2019-01-10
Cargo: remove tilde requirements of protobuf

overvenus:
So users can select protobuf version from "2.0.0" to "3.0.0". 

Cc #272


overvenus:
There was a failure caused by https://github.com/rust-lang/rustup.rs/issues/1609, but now it is fixed.

PR #275, opened 2019-01-21
 Bump v0.4.3

overvenus:
Remove tilde requirements `~2.0` of protobuf.

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -12,11 +12,11 @@ categories = ["network-programming"]
 build = "build.rs"
 
 [dependencies]
-protobuf = "~2.1"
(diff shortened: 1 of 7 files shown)


proto/Cargo.toml
```diff
@@ -12,11 +12,11 @@ categories = ["network-programming"]
 build = "build.rs"
 
 [dependencies]
-protobuf = "~2.0"
+protobuf = "~2.2"
```
messense: Why isn't this tilde removed?

overvenus: Because CI uses `grpcio-proto`, we dont want to break CI due to third party crates.

Issue #274, opened 2019-01-15
grpc-sys 0.4.2 fails to build - but 0.4.1 builds fine


richardwhiuk:
**Describe the bug**

0.4.2 fails to build for us with CentOS 7.

**To Reproduce**

On CentOS 7 run `cargo build` against a crate which depends on either grpc-sys 0.4.2, or grpc-sys 0.4.1

With grpc-sys 0.4.2, the following is seen:

```
error: failed to run custom build command for `grpcio-sys v0.4.2`
process didn't exit successfully: `/target/debug/build/grpcio-sys-840781b2e44e9c86/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-changed=grpc_wrap.cc
cargo:rerun-if-changed=grpc
cargo:rerun-if-env-changed=GRPCIO_SYS_USE_PKG_CONFIG
cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS
running: "cmake" "/opt/rust/registry/src/github.com-1ecc6299db9ec823/grpcio-sys-0.4.2/grpc" "-DGO_EXECUTABLE=fake-go-nonexist" "-DgRPC_INSTALL=false" "-DgRPC_BUILD_CSHARP_EXT=false" "-DgRPC_BUILD_CODEGEN=false" "-DCMAKE_INSTALL_PREFIX=.../target/debug/build/grpcio-sys-f45f9c164b4a8726/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_C_COMPILER=/usr/local/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_BUILD_TYPE=Debug"
-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
-- Check for working C compiler: /usr/local/bin/cc
-- Check for working C compiler: /usr/local/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
[...]

PR #278, opened 2019-01-29
Add binutils version

ice1000:
Signed-off-by: ice1000 <ice1000kotlin@foxmail.com>

See #274

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ This project is still under development. The following features with the check m
 
 - CMake >= 3.8.0
 - Rust >= 1.19.0
+- binutils >= 2.22
 - By default, the [secure feature](#feature-secure) is enabled, therefore Go (>=1.7) is required.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,6 +27,7 @@ This project is still under development. The following features with the check m
 
 - CMake >= 3.8.0
 - Rust >= 1.19.0
+- binutils >= 2.22
```
BusyJay: I don't think it's necessary. And it's usually a dependency of gcc.

ice1000: How about requiring a newer version of gcc?
Actually I don't know the corresponding gcc version.

BusyJay: Sounds more reasonable to me. Whatever works for grpc/grpc.

ice1000: According to https://grpc.io/about/

gcc **4.4, 4.6**, 5.3 are supported while in #274 people are using gcc 4.8 and 4.9.
It's like the version of binutils is independent to gcc.

PR #287, opened 2019-02-13
Add get_connectivity_state to Channel

zach-youssef:
Exposes the channel connectivity state struct and getter method from grpc::sys to the channel implementation.


src/channel.rs
```diff
@@ -562,6 +573,10 @@ impl Channel {
         }
     }
 
+    pub fn check_connectivity_state(&self, try_to_connect: bool) -> ConnectivityState {
```
BusyJay: Better add some comment to explain the side effect of the function.

src/channel.rs
```diff
@@ -531,6 +533,15 @@ struct ChannelInner {
     channel: *mut GrpcChannel,
 }
 
+impl ChannelInner{
+    pub(crate) fn check_connectivity_state(&self, try_to_connect: bool) -> ConnectivityState {
```
BusyJay: It's OK to keep the implementation in the function at L576.

zach-youssef:
Would you prefer for me to squash my fix commit or leave as-is?

Hoverbear:
@zach-youssef When the PR is merged we'll squash. Adding them as-is is convenient for code review.

Hoverbear:
@zach-youssef A `cargo fmt` is required to get the build to pass. :(

Hoverbear:
@BusyJay I resolved it, PTAL.

PR #291, opened 2019-02-21
Upgrade to Rust 2018

brson:
There are two notable things about this PR:

- `async` is a reserved word, so I renamed the module to `not_sync`. Happy for any better ideas.
- I bumped the version in case this causes breakage with older toolchains, which I suspect it does.

Fixes https://github.com/pingcap/grpc-rs/issues/256

PTAL @ice1000


nrc:
Why not `r#async`?

brson:
> Why not `r#async`

I'm assuming people don't want to write raw idents. I'm happy to leave it though.

src/lib.rs
```diff
@@ -56,31 +54,32 @@ mod env;
 mod error;
 mod log_util;
 mod metadata;
+mod not_sync;
```
ice1000: So this rename is to avoid clashing with a potential keyword?

ice1000: Ah it's mentioned in the OP.
Resolving.

brson: @ice1000 Yes, `async` is a keyword in 2018 that otherwise needs to be written as `r#async` (raw identifier). @nrc above suggested to just use `r#async`, which is also reasonable (and the word "async" is certainly more sensible than "not_sync".

Do you have opinions about what this module should be called? Whether it should stay `r#async`?

brson:
@BusyJay @overvenus you've both done a lot of work on this codebase. I'm interested to know what you think of naming the `async` model in 2018: `r#async`, `not_sync`, or something else? `async` is reserved in 2018.

BusyJay:
I don't have strong opinion. But `r#async` is really weird. Maybe a little bit off the topic, I am considering opting out futures support as an optional feature, so the name may be changed in the future.

brson:
Nightly build has failed twice in a row on two different tests. The CI build seems to be in a different configuration than locally, as I see a bunch of "E0221 19:17:53.770547100   25688 socket_utils_common_posix.cc:299] setsockopt(TCP_USER_TIMEOUT) Protocol not available" messages that aren't in the CI logs.

ice1000:
@brson there's one unstable test. Just restart it until it works.

src/lib.rs
overvenus: What about `task` or `executor`? This module impls a custom future executor based on grpc.

brson: I changed it to `task` since it's shorter than `executor`. Good suggestion.

brson:
CI error was due to some missing rustfmt changes. This time it'll work 🤞

PR #294, opened 2019-03-06
[WIP] add server instrumentation for calls

fbandersnatch:
I want to enable user-pluggable functionality to run on every request. The goal is to use it for logging and monitoring e.g., I'll use it for request counts, request rates and processing time counters.
This commit has a simple no-op example to achieve this.

Usage: 
```
// Replace the use of `create_greeter` with `create_instrumented_greeter`
let service = helloworld_grpc::create_instrumented_greeter(GreeterService, MyServerInstrumenter);
```
Implement `trait ServerInstrumenter` with the needed instrumentation, as `MyServiceInstrumenter`.

WIP:
This is still work in progress, I'm open to feedback on this

Are there plans to implement such use cases in other ways?


BusyJay:
Nice job! We planed to do this long time ago but still not get the time to make it really happen. Really excited to see someone can pick it up finally!

The underlying architecture of grpcio is very similar to the C# implementation. So you can refer to the similar proposal and its implementation at grpc/proposal#38.

fbandersnatch:
Cool, I'll look at their implementation.
For now, I'm not sure why compilation is failing, I'll submit an update just to kick a build.

PR #300, opened 2019-04-03
Bump to 0.5.0

overvenus:
Update change log and bump to 0.5.0

### DIFF
--- a/grpc-sys/Cargo.toml
+++ b/grpc-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio-sys"
-version = "0.4.0"
+version = "0.5.0"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
 keywords = ["grpc", "bindings"]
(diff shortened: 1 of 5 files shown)


grpc-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio-sys"
-version = "0.4.0"
+version = "0.5.0"
```
nrc: I think we need to update the other Cargo tomls too? And should it be `0.5.0-preview`?

overvenus: We have updated other tomls to 0.5.0 on #256 . I am sure whether semver supports the `preview` suffix.

nrc: > We have updated other tomls to 0.5.0 on #256

Oh, great!

> I am sure whether semver supports the preview suffix.

It does (as do crates.io and Cargo): https://semver.org/#spec-item-9

nrc:
\o/

PR #309, opened 2019-04-05
update codegens to move the last instance instead of clone

abbychau:
#304


compiler/src/codegen.rs
```diff
@@ -567,4 +571,4 @@ pub fn gen(
 
 pub fn protoc_gen_grpc_rust_main() {
     compiler_plugin::plugin_main(gen);
-}
+}
```
ice1000: Please don't remove this line end.

compiler/src/prost_codegen.rs
```diff
@@ -535,4 +544,4 @@ fn generate_method_bind(service_name: &str, method: &Method, buf: &mut String) {
     buf.push_str(", move |ctx, req, resp| instance.");
     buf.push_str(&method.name);
     buf.push_str("(ctx, req, resp));\n");
-}
+}
```
ice1000: Ditto

ice1000:
It is important for you to pass CI tests. It's not possible to merge if CI is not happy (we do fmt check on CI).

compiler/src/codegen.rs
abbychau: Done

compiler/src/prost_codegen.rs
abbychau: Done

Issue #306, opened 2019-04-05
Unstable test: ping_pong::test_secure


nrc:
Only observed with `--features "openssl"`

nrc:
Haven't seen this for a while

Hoverbear:
@nrc Maybe you haven't seen it due to #310?

Issue #305, opened 2019-04-05
Unstable test: client_streaming::test_secure


nrc:
So far only observed with `--features "openssl"`

nrc:
Hitting this one a lot recently.

nrc:
I'm seeing this again in https://travis-ci.org/pingcap/grpc-rs/jobs/559681593 with `cargo test --features "openssl" --all`, which doesn't make sense, since it should be ignored after #310. I still can't reproduce locally (Mac OS) though :-(

PR #310, opened 2019-04-05
Ignore intermittent tests

nrc:
Ignore tests which do not reliably pass. Since everyone just restarts until they pass, they don't add much benefit and they waste resources and people's time.

Long-term it would be good to fix these tests, but until somebody does that I think it is better not to run them.


nrc:
Ironically, there is a test error in one of the Appveyor builds, but it looks like an infra failure rather than anything wrong with this PR.

nrc:
PTAL @siddontang @BusyJay @overvenus @Hoverbear Does anyone not approve of this approach to the unstable tests?

Hoverbear:
I worked at some old clients that had this approach and the moment you remove the pain, embarrassment, and reminders, everyone forgets about it and they never get fixed.

So, I'm generally of the opinion that the tests should get fixed (and be stable) or removed completely (because they are superseded with others).

nrc:
> the moment you remove the pain, embarrassment, and reminders, everyone forgets about it and they never get fixed.

I don't want to sound sarcastic, but they are not getting fixed right now, and they are a big dent in development velocity. It seems like wishful thinking to suppose they will get fixed by leaving things unchanged.

The downside of the current situation is that we are so used to re-running tests that tests could change from passing to intermittent without people noticing. If the default is that all tests pass, then that doesn't happen.

I hope that by keeping track of the number of intermittent failures and keeping that number highlighted (e.g., by making a tracking issue and pinning it, or using a dashboard), we can encourage the tests to get fixed, even without the pain.

ice1000:
@Hoverbear Notice that these tests can be seen as completely removed already, but they're only removed for some certain features. Does that sound reasonable enough?

BusyJay:
This is too bad that disable the test cases instead of fixing them. It's called 掩耳盗铃 in Chinese that cover one's ears when stealing a bell.

@ice1000 You should not merge a PR when there is only one approval. Ask someone to review the code instead.

ice1000:
@BusyJay I recommend requiring two approvals for merging. There's only on required for this repo atm so I assumed that it's fine to merge with 1 approval.

Btw there are issues for the disabled tests: #305 #306 so what we're doing is treating the disabling of the tests as a "bug" and postpone the fix.
I'm fine with regression, because they're disabled for not blocking some PRs and the blocking PRs are mostly merged now.

Hoverbear:
I think all repos except some trivial documentation changes have always required 2 approvals in all PingCAP and TiKV repos.

When there are folks who are still actively against the merging of the PR it's probably good form to wait for a second approval.

PR #311, opened 2019-04-06
 make `./proto` buildable

abbychau:
it is not buildable and propagating to `interop` due to the wrong path of `example`


ice1000:
What do you mean by build `./proto`? What's wrong before?

ice1000:
Why are the two .desc files added?

abbychau:
https://github.com/pingcap/grpc-rs/blob/b2088436969f2ec248086b91b1c35bad59042b33/proto/build.rs#L135
```Rust
fn main() {
    for package in &["testing", "example", "health"] {
        compile_protobuf(format!("{}.desc", package), package);
        compile_prost(format!("{}.desc", package));
    }
}

```
is requiring generated desc from `example` , where `compile_protobuf` are `desc_to_module_prost` unwrapping (assuming existence)


those files are apparently generating by https://github.com/pingcap/grpc-rs/blob/b2088436969f2ec248086b91b1c35bad59042b33/proto/build.sh

but the command `example example example` is wrong.

ice1000:
> is requiring generated desc from example , where compile_protobuf are desc_to_module_prost unwrapping (assuming existence)

Could you please run the script to generate those .desc files instead of adding them to this git repo?

abbychau:
> > is requiring generated desc from example , where compile_protobuf are desc_to_module_prost unwrapping (assuming existence)
> 
> Could you please run the script in build.rs to generate those .desc files instead of adding them to this git repo?

I actually think that it is a better way, but I also notice that there is a member called jay committed .desc files for the other two folder of protos, so I thought that it could be a thought decision.

besides, there is no clear documentations or triggers to run `build.sh`.


Do you suggest to put a shell execution pointing `build.sh` inside build.rs?
I am not very sure if it is a good one because it could make it not portable to windows.

ice1000:
> I actually think that it is a better way, but I also notice that there is a member called jay committed .desc files for the other two folder of protos, so I thought that it could be a thought decision.

This is at least reasonable to me. GJ!

PR #323, opened 2019-05-20
Fix Environment thread leak.

smoelius:
`EnvBuilder::build` spawns threads, but those threads are never joined.  This becomes a problem, say, when you fuzz an application that calls `EnvBuilder::build`.  More specifically, in long runs of `cargo-fuzz`, AddressSanitizer will complain that too many threads have been created.  This pull request has the `Environment` join with the threads in its `drop` method.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -140,6 +140,10 @@ impl Drop for Environment {
             // it's safe to shutdown more than once.
             cq.shutdown()
         }
+
+        for handle in self._handles.drain(..) {
+            handle.join().unwrap();
+        }
     }
 }
 


src/env.rs
```diff
@@ -140,6 +140,10 @@ impl Drop for Environment {
             // it's safe to shutdown more than once.
             cq.shutdown()
         }
+
+        for handle in self._handles.drain(..) {
+            handle.join().unwrap();
```
ice1000: How about error handling for joining the created threads? Are there potential panics/latencies?

BusyJay:
Actually when the queue is shutdown, the threads will exit automatically in the end. What threads did you see in your tests? Does this PR solve your problem?

smoelius:
While trying to produce a minimal example, I realized that this PR does not in fact solve my problem.

It appears now that this may be a limitation of AddressSanitizer.

Sorry for the noise.

ice1000:
Don't worry, but thanks for telling us this limitation.

PR #326, opened 2019-05-30
grpcio-sys: fix openssl vendor build

BusyJay:
Vendor build implemented in #262 is not configured properly, compiler
will use the system library during compilation. This PR fixes the
problem by configuring package searching path for openssl.

### DIFF
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -69,6 +69,7 @@ fn build_grpc(cc: &mut Build, library: &str) {
 
     let dst = {
         let mut config = Config::new("grpc");
+
         if !cfg!(feature = "secure") {
             // boringssl's configuration is still included, but targets
             // will never be built, hence specify a fake go to get rid of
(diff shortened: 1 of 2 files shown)


grpc-sys/build.rs
```diff
@@ -203,14 +203,21 @@ fn build_grpc(cc: &mut Build, library: &str) {
...
-    );
-    println!("cargo:include={}", include_dir.to_string_lossy());
+    // openssl-sys uses openssl-src to build the library. openssl-src uses
+    // configure/make to build the library which makes it hard to detect
+    // what's the actual path of the library. Here assumes the directory
+    // structure as follow (which is the behavior of 0.9.47):
+    // install_dir/
+    //     include/
+    //     lib/
+    // Remove the hack when sfackler/rust-openssl#1117 is resolved.
```
ice1000: Should we document this?

BusyJay: The structure is unlikely to be changed, so I think comment is OK.

ice1000:
Nightly CI failing due to linter errors: https://travis-ci.org/pingcap/grpc-rs/jobs/539159730

BusyJay:
I'm going to merge this. The lint problem should be fixed in other PR.

PR #332, opened 2019-06-19
grpcio-sys: add bindgen feature

hunterlxt:
Add bindgen feature into `build.rs` and fix some interfaces in the downstream repo. 
To be compatible with the previous code design, a new `grpc-wrap.rs` was created.
The ultimate goal is to rewrite all `grpc_wrap.cc` with rust, but now only bindgen is generated for wrap.


hunterlxt:
All checks have failed, It seems that CI lacks package `Clang` for `bindgen`.

grpc-sys/build.rs
```diff
@@ -234,7 +237,78 @@ fn get_env(name: &str) -> Option<String> {
...
+    let mut config = config_from_headers();
+
+    if get_env("CARGO_CFG_TARGET_OS").map_or(false, |s| s == "windows") {
+        // At lease win7
+        config = config.clang_arg("-D _WIN32_WINNT=0x0700");
+    }
+
+    let bindings = config
+        .clang_arg("-I./grpc/include")
+        .clang_arg("-DGRPC_SYS_SECURE")
```
BusyJay: Why define it? What if secure feature is disabled?

grpc-sys/build.rs
```diff
@@ -234,7 +237,78 @@ fn get_env(name: &str) -> Option<String> {
...
+fn config_from_headers() -> bindgen::Builder {
+    let mut config = bindgen::Builder::default();
+    let mut headers: Vec<String> = Vec::new();
+    let matcher = RegexMatcher::new("GRPCAPI|GPRAPI").expect("can't resolve pattern");
+
+    for result in WalkDir::new(Path::new("./grpc/include")) {
+        let dent = match result {
+            Ok(dent) => dent,
+            Err(err) => {
+                eprintln!("{}", err);
```
BusyJay: I think it should panic to avoid missing headers.

grpc-sys/build.rs
```diff
@@ -234,7 +237,78 @@ fn get_env(name: &str) -> Option<String> {
...
+                continue;
+            }
+        };
+        if !dent.file_type().is_file() {
+            continue;
+        }
+        let result = grep::searcher::Searcher::new().search_path(
```
BusyJay: Can you use pure string comparison instead?

grpc-sys/src/grpc_wrap.rs
```diff
@@ -0,0 +1,120 @@
...
+            8 => grpc_status_code::GRPC_STATUS_RESOURCE_EXHAUSTED,
+            9 => grpc_status_code::GRPC_STATUS_FAILED_PRECONDITION,
+            10 => grpc_status_code::GRPC_STATUS_ABORTED,
+            11 => grpc_status_code::GRPC_STATUS_OUT_OF_RANGE,
+            12 => grpc_status_code::GRPC_STATUS_UNIMPLEMENTED,
+            13 => grpc_status_code::GRPC_STATUS_INTERNAL,
+            14 => grpc_status_code::GRPC_STATUS_UNAVAILABLE,
+            15 => grpc_status_code::GRPC_STATUS_DATA_LOSS,
+            16 => grpc_status_code::GRPC_STATUS_UNAUTHENTICATED,
+            _ => panic!("get wrong grpc_status_code"),
```
BusyJay: Why not just convert to `GRPC_STATUS_UNKNOWN` just like before?

BusyJay:
I suggest to provide a new StatusCode type in grpcio library to make it more rusty and remove annoying DO_NOT_USE fields. And as the name says, maybe we should just define StatusCode as a struct `StatusCode(u64)` instead of an enum.

grpc-sys/build.rs
hunterlxt: see [here](https://github.com/pingcap/grpc-rs/blob/master/grpc-sys/grpc_wrap.cc#L782) if the feature is disabled, LLVM can't compile those code > <

grpc-sys/build.rs
hunterlxt: solved !

grpc-sys/build.rs
hunterlxt: Did not fully understand...

grpc-sys/src/grpc_wrap.rs
hunterlxt: In gRPC, code `2` represents UNKNOWN, so I think when you receive a num that is not in scope, something must be wrong. Unfortunately, this will panic, so if it's better to set it into UNKNOWN, I can reverse it.

grpc-sys/src/grpc_wrap.rs
```diff
@@ -93,28 +93,31 @@ impl grpc_byte_buffer_reader {
     }
 }
 
-impl From<i32> for grpc_status_code {
-    fn from(code: i32) -> grpc_status_code {
+
+pub struct GrpcStatusCode(u64);
```
ice1000: How about `GrpcStatusCode(pub u64)`?

hunterlxt: > Please don't use Chinese in your comments

Forget it, just a test...

hunterlxt:
After discussion with @BusyJay , `StatusCode` is better to use i32. Considering this code may add custom information, so it is not made into rust enum.

grpc-sys/build.rs
BusyJay: It's intended not to compile those code when secure feature is disabled. See also https://github.com/pingcap/grpc-rs/blob/d00a8370183a4f0057378cf10cae8476c2dffe64/grpc-sys/build.rs#L303

grpc-sys/src/grpc_wrap.rs
ice1000: Yes but that's gonna get stored in GitHub jquery cache for some sense (in case you can delete them)

hunterlxt: Covered by `--force`

grpc-sys/build.rs
hunterlxt: solved

grpc-sys/src/grpc_wrap.rs
ice1000: :+1:

grpc-sys/src/grpc_wrap.rs
```diff
@@ -0,0 +1,94 @@
+use super::*;
+
+use std::slice;
+use std::time::Duration;
+
+pub const GRPC_INITIAL_METADATA_IDEMPOTENT_REQUEST: u32 = 0x0000_0010;
```
BusyJay: Aren't these consts part of generated bindings?

src/call/mod.rs
```diff
@@ -31,9 +31,10 @@ use crate::codec::{DeserializeFn, Marshaller, SerializeFn};
 use crate::error::{Error, Result};
 use crate::task::{self, BatchFuture, BatchType, CallTag, SpinLock};
 
-pub use crate::grpc_sys::GrpcStatusCode as RpcStatusCode;
+pub use crate::grpc_sys::grpc_status_code as RpcStatusCode;
+pub type RpcStatusCodeType = i32;
```
BusyJay: Why not use `RpcStatusCode` directly?

hunterlxt: In this case, `RpcStatusCode` is a module name, bindgen generates those const in that module. To represent the type of the code, we need a code-type-name.

grpc-sys/src/grpc_wrap.rs
hunterlxt: Solved

PR #334, opened 2019-06-21
grpcio-sys: fix warnings in nightly

hunterlxt:
1. error: trait objects without an explicit `dyn` are deprecated
2. `Box<trait>` to `Box<dyb trait>`
3. libc has deprecated several types, which are used in grpcio.

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -19,10 +19,11 @@ bytes = "0.4"
 prost = "0.5"
 prost-derive = "0.5"
 protobuf = "~2.2"
+lazy_static = "1.3"
 
 [build-dependencies]
 grpcio-compiler = { path = "../compiler" }
-protobuf-build = "0.2"
+protobuf-build = "0.6"
 prost = "0.5"
 prost-types = "0.5"
 protobuf = "~2.2"
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -23,7 +23,7 @@ lazy_static = "1.3"
 
 [build-dependencies]
 grpcio-compiler = { path = "../compiler" }
-protobuf-build = "0.6.2"
+protobuf-build = "0.6"
 prost = "0.5"
 prost-types = "0.5"
 protobuf = "~2.2"
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -23,7 +23,7 @@ lazy_static = "1.3"
 
 [build-dependencies]
 grpcio-compiler = { path = "../compiler" }
-protobuf-build = "0.6"
+protobuf-build = "0.6.2"
 prost = "0.5"
 prost-types = "0.5"
 protobuf = "~2.2"
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -11,8 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[allow(dead_code)]
-#[allow(clippy::all)]
+#[allow(clippy::large_enum_variant)]
 pub mod testing {
     include!(concat!(env!("OUT_DIR"), "/grpc.testing.rs"));
     include!(concat!(env!("OUT_DIR"), "/wrapper_grpc.testing.rs"));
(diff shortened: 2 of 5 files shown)


proto/src/lib.rs
```diff
@@ -13,6 +13,7 @@
 
 #[allow(dead_code)]
 #[allow(clippy::all)]
+#[allow(bare_trait_objects)]
```
BusyJay: Can you fix it in the grpcio-compiler?

hunterlxt: I found related code generation is here: https://github.com/tikv/protobuf-build
This fix work may require the help of @nrc

hunterlxt: @BusyJay I found related code generation is here: https://github.com/tikv/protobuf-build, not in codegen.rs in grpc-compiler.
So this fix work may require the help of @nrc

ice1000: Have you tried updating the protobuf-build dependency and running a re-generation?

hunterlxt: > Have you tried updating the protobuf-build dependency and running a re-generation?

Thanks for suggestion, I will try tomorrow.

hunterlxt:
@nrc How long does it take to update protobuf-build?

nrc:
> @nrc How long does it take to update protobuf-build?

You just need to change the version in Cargo.toml, then run with `--features gen`

hunterlxt:
> > @nrc How long does it take to update protobuf-build?
> 
> You just need to change the version in Cargo.toml, then run with `--features gen`

It can't work, no features `gen` found in grpc-rs.

In order to advance the progress, this PR needs to be merged. @BusyJay

nrc:
@hunterlxt sorry, I mis-remembered, there is no `gen` feature in this project. You just need to update protobuf-build. The diff for doing that is:

```
diff --git a/proto/Cargo.toml b/proto/Cargo.toml
index 6a0766f..1a5e532 100644
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -19,10 +19,11 @@ bytes = "0.4"
 prost = "0.5"
 prost-derive = "0.5"
 protobuf = "~2.2"
+lazy_static = "1.3"

 [build-dependencies]
 grpcio-compiler = { path = "../compiler" }
-protobuf-build = "0.2"
+protobuf-build = "0.6"
 prost = "0.5"
 prost-types = "0.5"
 protobuf = "~2.2"
diff --git a/proto/build.rs b/proto/build.rs
index 379cc20..3480852 100644
--- a/proto/build.rs
+++ b/proto/build.rs
@@ -92,6 +92,7 @@ fn desc_to_module_prost(descriptor: &Path) {
             .map(|m| format!("{}/{}.rs", env::var("OUT_DIR").unwrap(), m))
             .collect::<Vec<_>>(),
         &env::var("OUT_DIR").unwrap(),
+        protobuf_build::GenOpt::all(),
     );
 }

```

proto/Cargo.toml
```diff
@@ -19,10 +19,11 @@ bytes = "0.4"
 prost = "0.5"
 prost-derive = "0.5"
 protobuf = "~2.2"
+lazy_static = "1.3"
 
 [build-dependencies]
 grpcio-compiler = { path = "../compiler" }
-protobuf-build = "0.2"
+protobuf-build = "0.6.2"
```
nrc: Better to just use "0.6" here. Then we'll get more improvements without having to land changes.

hunterlxt: I try "0.6" before, but got some warnings. Hmmm... you told me that 0.6.2 fixed this problem.

nrc: Version 0.6.2 is the version of the crate with the fix. Using "0.6" means Cargo will choose from 0.6.0, 0.6.1, 0.6.2, and in the future, 0.6.3, etc. If you use just "0.6" it should be fine, you might need to `cargo update -p protobuf-build`, but I don't think you'll need to.

hunterlxt: understood, fixed

BusyJay: Choosing 0.6.2  should mean any version that is >=0.6.2 && < 0.7.0 , so it can still get further improvements.

BusyJay:
Good job!

PR #337, opened 2019-06-26
grpcio: update travis

### LINKED ISSUE #335: Update travis configuration
Travis is used as a tool for continuous integration in many open source projects. It is simple and easy to use, and there are more users, which means that some problems can easily be solved. Travis used to not support the Windows platform, we maintain two CI tools.

1. I'm considering moving Windows testing part to Travis. This doesn't look hard
2. Bindgen need LLVM, the package need to be declared in `.travis.yml`

Why Bindgen? See [here](https://github.com/pingcap/grpc-rs/issues/331)

hunterlxt:
1. Add Windows support
2. update `llvm` dependency for `bindgen`

Related issue #335

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,54 @@
-dist: trusty
+dist: xenial
 sudo: false
 language: rust
-cache:
-  directories:
-    - $HOME/.cargo
-    - $HOME/.cache
-    - $TRAVIS_BUILD_DIR/target
-before_cache:
-  - find $TRAVIS_BUILD_DIR/target/debug -maxdepth 1 -type f -delete
+cache: cargo
 
 env:
   global:
     - RUST_BACKTRACE=1
     - RUSTFLAGS="--deny=warnings"
 
-os:
-  - linux
-rust:
-  - stable
-  - beta
-  - nightly
 matrix:
   include:
-  - os: osx
+  - name: linux-stable
+    rust: stable
+    addons:
+      apt:
+        sources:
+          - llvm-toolchain-xenial-7
+        packages:
+          - clang-tidy
+          - llvm-dev
+          - libclang-dev
+          - clang
+
+  - name: linux-nightly
+    rust: nightly
+    addons:
+      apt:
+        sources:
+          - llvm-toolchain-xenial-7
+        packages:
+          - clang-tidy
+          - llvm-dev
+          - libclang-dev
+          - clang
+
+  - name: win-stable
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,28 +1,54 @@
...
 sudo: false
 language: rust
-cache:
-  directories:
-    - $HOME/.cargo
-    - $HOME/.cache
-    - $TRAVIS_BUILD_DIR/target
-before_cache:
-  - find $TRAVIS_BUILD_DIR/target/debug -maxdepth 1 -type f -delete
+cache: cargo
```
BusyJay: Does it include `$HOME/.cache`?

.travis.yml
```diff
@@ -1,28 +1,54 @@
-dist: trusty
+dist: xenial
 sudo: false
 language: rust
-cache:
-  directories:
-    - $HOME/.cargo
-    - $HOME/.cache
-    - $TRAVIS_BUILD_DIR/target
-before_cache:
```
BusyJay: The step is necessary to keep cache small.

.travis.yml
```diff
@@ -1,28 +1,54 @@
...
-  - nightly
 matrix:
   include:
-  - os: osx
+  - name: linux-stable
+    rust: stable
+    addons:
+      apt:
+        sources:
+          - llvm-toolchain-xenial-7
```
BusyJay: Why repeat everything? What's wrong with the original configurations?

.travis.yml
hunterlxt: Because the Windows platform was only supported at last year, I did not find useful information about the cache in Windows platform.
For insurance, just cache the cargo directory.
After testing, the cache seems to have little effect on speed, but I can restore these deleted items and leave them unchanged if necessary.

.travis.yml
hunterlxt: I found that the three platforms are very different.
For easy understanding and configuration convenience
If don't write into matrix, I have encountered many problems after introducing Windows platform.

hunterlxt: However, the two Linux platform does not need to be repeated.

.travis.yml
hunterlxt: even on windows platform ? maybe i need more test tonight

hunterlxt:
> Are we going to have win/osx nightly builds?

Linux is the primary support platform, and does not consider nightly testing of all platforms for the time being.

PR #343, opened 2019-07-16
Use feature flags for Prost/protobuf in proto and compiler

nrc:
Let the protos and compiler depend on either Prost or rust-protobuf without needing both as deps (note that the compatibility layer for Prost is based on the `protobuf::Message` trait, so we still need the `protobuf` dep in the Prost case). Fixes some slightly silly code in proto::utils

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ path = "examples/hello_world/server.rs"
 serde_json = "1.0"
 serde = "1.0"
 serde_derive = "1.0"
-grpcio-proto = { path = "proto", version = "0.5.0-alpha" }
+grpcio-proto = { path = "proto", features = ["prost-codec"], default-features = false }
 rand = "0.4"
 slog = "2.0"
 slog-async = "2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ path = "examples/hello_world/server.rs"
 serde_json = "1.0"
 serde = "1.0"
 serde_derive = "1.0"
-grpcio-proto = { path = "proto", features = ["prost-codec"], default-features = false }
+grpcio-proto = { path = "proto", default-features = false }
 rand = "0.4"
 slog = "2.0"
 slog-async = "2.1"
--- a/compiler/Cargo.toml
+++ b/compiler/Cargo.toml
@@ -11,14 +11,20 @@ documentation = "https://docs.rs/grpcio-compiler"
 description = "gRPC compiler for grpcio"
 categories = ["network-programming"]
 
+[features]
+default = ["protobuf-codec"]
+protobuf-codec = ["protobuf-codegen"]
+prost-codec = ["prost-build", "prost-types", "prost"]
+
 [dependencies]
-protobuf = "2.0"
-protobuf-codegen = "2.0"
-prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-prost-build = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-prost-types = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
+protobuf = "2"
+prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
+prost-build = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
+prost-types = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
+protobuf-codegen = { version = "2", optional = true }
 derive-new = "0.5"
 tempfile = "3.0"
 
 [[bin]]
 name = "grpc_rust_plugin"
+required-features = ["protobuf-codec"]
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -14,8 +14,8 @@ build = "build.rs"
 
 [features]
 default = ["protobuf-codec"]
-protobuf-codec = ["grpcio/protobuf-codec", "protobuf-codegen", "protobuf"]
-prost-codec = ["prost-derive", "bytes", "lazy_static", "grpcio/prost-codec", "prost-types", "prost"]
+protobuf-codec = ["grpcio/protobuf-codec", "protobuf-codegen", "protobuf", "grpcio-compiler/protobuf-codec"]
+prost-codec = ["prost-derive", "bytes", "lazy_static", "grpcio/prost-codec", "prost-types", "prost", "grpcio-compiler/prost-codec"]
 
 [dependencies]
 futures = "0.1"
@@ -28,8 +28,7 @@ lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
 protobuf-build = "0.7"
-grpcio-compiler = { path = "../compiler" }
-protobuf-build = "=0.6.2"
+grpcio-compiler = { path = "../compiler", default-features = false }
 prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
 prost-types = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
 protobuf = { version = "2", optional = true }
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -12,19 +12,25 @@ description = "Public proto files for grpcio."
 categories = ["network-programming"]
 build = "build.rs"
 
+[features]
+default = ["protobuf-codec"]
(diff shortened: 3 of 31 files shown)


proto/Cargo.toml
```diff
@@ -12,19 +12,24 @@ description = "Public proto files for grpcio."
...
+prost-derive = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
+protobuf = "2"
+lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
-grpcio-compiler = { path = "../compiler" }
-protobuf-build = "=0.6.2"
-prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-prost-types = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-protobuf = "~2.2"
```
Hoverbear: 👍 This is an important change for later

proto/build.rs
```diff
@@ -11,131 +11,151 @@
...
     }
+}
 
-    for (name, content) in results {
-        let out_file = output_dir.join(&name);
-        let mut f = File::create(&out_file).unwrap();
-        f.write_all(&content).unwrap();
-        let (module_name, _) = name.split_at(name.len() - 3); // ".rs".len() == 3
-        writeln!(out, "pub mod {};", module_name).unwrap();
+#[cfg(feature = "prost-codec")]
```
Hoverbear: Why do you opt for this over `cfg(feature ...)`'ing in the `main()`?

proto/src/lib.rs
```diff
@@ -133,4 +136,27 @@ pub mod health {
     }
 }
 
+#[cfg(all(feature = "protobuf-codec", not(feature = "prost-codec")))]
```
Hoverbear: Why does these ones have a more complicated `cfg` block compared to the above? Is it related to defaults?

proto/build.rs
nrc: because that alternative means the functions must compile even if they are not used, which means that if using the protobuf feature, we would still need to depend on Prost

proto/src/lib.rs
nrc: It is because Cargo can arbitrarily combine features so we must be able to support both features at the same time. If we use the simpler cfg here, then we would get two versions of the same name in the module.

proto/build.rs
```diff
@@ -11,131 +11,151 @@
...
-            .map(|m| format!("{}/{}.rs", env::var("OUT_DIR").unwrap(), m))
-            .collect::<Vec<_>>(),
-        &env::var("OUT_DIR").unwrap(),
-        protobuf_build::GenOpt::all(),
-    );
+#[cfg(not(feature = "prost-codec"))]
+mod prost {
+    use std::path::Path;
+
+    pub fn compile<P: AsRef<Path>>(_desc_path: P) {}
```
ice1000: Why is this needed?

Hoverbear: I think it's related to https://github.com/pingcap/grpc-rs/pull/343#discussion_r304208236

nrc: Yeah, it gets called from `main` so you need the stub function here. Alternative would be to `cfg` the call in main, I don't mind either version tbh

Cargo.toml
```diff
@@ -57,7 +57,7 @@ path = "examples/hello_world/server.rs"
 serde_json = "1.0"
 serde = "1.0"
 serde_derive = "1.0"
-grpcio-proto = { path = "proto", version = "0.5.0-alpha" }
+grpcio-proto = { path = "proto", features = ["prost-codec"], default-features = false }
```
BusyJay: Why use prost-codec features?

compiler/Cargo.toml
```diff
@@ -11,14 +11,20 @@ documentation = "https://docs.rs/grpcio-compiler"
...
+prost-codec = ["prost-build", "prost-types", "prost"]
+
 [dependencies]
-protobuf = "2.0"
-protobuf-codegen = "2.0"
-prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-prost-build = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-prost-types = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
+protobuf = "2"
+prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77", optional = true }
```
BusyJay: Can 0.5 be used?

proto/Cargo.toml
```diff
@@ -12,19 +12,24 @@ description = "Public proto files for grpcio."
...
 build = "build.rs"
 
+[features]
+default = ["protobuf-codec"]
+protobuf-codec = ["grpcio/protobuf-codec", "protobuf-codegen", "protobuf", "grpcio-compiler/protobuf-codec"]
+prost-codec = ["prost-derive", "bytes", "lazy_static", "grpcio/prost-codec", "prost-types", "prost", "grpcio-compiler/prost-codec"]
+
 [dependencies]
 futures = "0.1"
 grpcio = { path = "..", features = ["prost-codec"] }
```
BusyJay: So by default, when depends on grpcio-proto, both prost-codec and protobuf-codec will be enabled?

Cargo.toml
nrc: It was easiest (because this what was done already)? I think it could support both, but I think for an example it is maybe OK to pick one.

compiler/Cargo.toml
nrc: 0.5 is missing some optimisations and a recursion limit. I think we need to have a 0.6 release before we can move to Prost properly (or even do useful benchmarking), but we could stick to 0.5 if you think it is important.

proto/Cargo.toml
nrc: whoops, I missed that one. Addressed in the latest push

Cargo.toml
BusyJay: Then how can we know our default codecs work as expected if it's not being used by any crates?

proto/build.rs
```diff
@@ -11,131 +11,151 @@
...
-fn write_files<W: Write>(
-    results: impl Iterator<Item = (String, Vec<u8>)>,
-    output_dir: &Path,
-    mut out: W,
-) {
-    if !output_dir.exists() {
-        fs::create_dir_all(output_dir).unwrap();
+fn main() {
+    for package in &["testing", "example", "health"] {
+        protobuf::compile(format!("{}.desc", package), package);
```
BusyJay: Why not use one line instead? So only one of prost-codec and protobuf-codec will be activated.

proto/src/lib.rs
```diff
@@ -133,4 +136,27 @@ pub mod health {
...
+#[allow(bare_trait_objects)]
+#[allow(renamed_and_removed_lints)]
+pub mod testing {
+    include!(concat!(env!("OUT_DIR"), "/testing/mod.rs"));
+}
+
+#[cfg(all(feature = "protobuf-codec", not(feature = "prost-codec")))]
```
BusyJay: I think only one feature should be activated.

Cargo.toml
nrc: Fair. We can probably use both codecs

proto/build.rs
nrc: Hmm, maybe could, we need to support generating for both codecs, but I think because of the complex cfgs, we'll only use one anyway...

proto/src/lib.rs
nrc: Unfortunately Cargo will build with multiple features, even if they're not usually used together. This happened when running `cargo test --all`. Unfortunately Cargo doesn't have any way of finding out why certain features are used. That might be due to the mistake in the Cargo.toml you noted above. I'll try it out....

Cargo.toml
BusyJay: I think we can just use rust-protobuf for now as first step.

proto/build.rs
BusyJay: Use only one line can only generate what's necessary. And following the modules can also be simplified.

proto/build.rs
ice1000: Ok

nrc:
@BusyJay I think I've addressed all your comments in the last two commits. PTAL.

proto/Cargo.toml
```diff
@@ -12,19 +12,24 @@ description = "Public proto files for grpcio."
...
+
 [dependencies]
 futures = "0.1"
-grpcio = { path = "..", features = ["prost-codec"] }
-bytes = "0.4"
-prost = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-prost-derive = { git = "https://github.com/danburkert/prost", rev = "1944c27c3029d01ff216e7b126d846b6cf8c7d77" }
-protobuf = "~2.2"
-lazy_static = "1.3"
+grpcio = { path = "..", features = ["secure"], default-features = false }
```
BusyJay: Why proto requires secure features?

BusyJay: Turns out proto needs to generate credentials, so secure feature is necessary.

PR #351, opened 2019-07-24
*: fix builds

BusyJay:
This PR breaks the circle dependencies between grpcio and grpcio-proto
by adding a new crate named tests and examples. We can bring things back
once rust-lang/cargo#6915 is fixed.

The new structure also fixes a problem that crates.io denies to accept
circle dependencies in dev-dependencies. More see rust-lang/cargo#4242.

This PR also updates protobuf-build to the latest version.


proto/Cargo.toml
```diff
@@ -27,9 +27,10 @@ protobuf = "2"
 lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
-protobuf-build = "0.7"
+protobuf-build = { version = "0.8", default-features = false }
 grpcio-compiler = { path = "../compiler", version = "0.5.0-alpha.2", default-features = false }
 prost = { version = "0.5", optional = true }
 prost-types = { version = "0.5", optional = true }
 protobuf = { version = "2", optional = true }
```
nrc: I think all these optional build deps can be removed

BusyJay: You are right.

tests-and-examples/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
overvenus: Could you leave a comment about the purpose of the empty file?

overvenus:
Should we run tests with the `prost-codec`?

BusyJay:
No, due to the difference mechanism between packages and modules, there are still things to do to support run tests against it.

hunterlxt:
seems that CI in Win platform met some errors. We may ignore it ?

BusyJay:
It seems tortall is down. If all the other CI passes, I think we can merge it.

PR #354, opened 2019-07-24
grpcio: wrap the grpc status code

hunterlxt:
Give the variable a more descriptive name so we know what kind of a code it is and what the valid values are. Leave grpc-sys untouch and define the type `RpcStatusCode` in grpcio itself.

### DIFF
--- a/interop/src/server.rs
+++ b/interop/src/server.rs
@@ -60,7 +60,7 @@ impl TestService for InteropTestService {
         if req.has_response_status() {
             let code = req.get_response_status().get_code();
             let msg = Some(req.take_response_status().take_message());
-            let status = RpcStatus::new(RpcStatusCode::new(code), msg);
+            let status = RpcStatus::new(code.into(), msg);
             let f = sink
                 .fail(status)
                 .map_err(|e| panic!("failed to send response: {:?}", e));
@@ -139,7 +139,7 @@ impl TestService for InteropTestService {
                 if req.has_response_status() {
                     let code = req.get_response_status().get_code();
                     let msg = Some(req.take_response_status().take_message());
-                    let status = RpcStatus::new(RpcStatusCode::new(code), msg);
+                    let status = RpcStatus::new(code.into(), msg);
                     failure = Some(sink.fail(status));
                 } else {
                     let mut resp = StreamingOutputCallResponse::default();
--- a/interop/src/server.rs
+++ b/interop/src/server.rs
@@ -60,7 +60,7 @@ impl TestService for InteropTestService {
         if req.has_response_status() {
             let code = req.get_response_status().get_code();
             let msg = Some(req.take_response_status().take_message());
-            let status = RpcStatus::new(code.into(), msg);
+            let status = RpcStatus::new(code, msg);
             let f = sink
                 .fail(status)
                 .map_err(|e| panic!("failed to send response: {:?}", e));
@@ -139,7 +139,7 @@ impl TestService for InteropTestService {
                 if req.has_response_status() {
                     let code = req.get_response_status().get_code();
                     let msg = Some(req.take_response_status().take_message());
-                    let status = RpcStatus::new(code.into(), msg);
+                    let status = RpcStatus::new(code, msg);
                     failure = Some(sink.fail(status));
                 } else {
                     let mut resp = StreamingOutputCallResponse::default();
--- a/interop/src/server.rs
+++ b/interop/src/server.rs
@@ -60,7 +60,7 @@ impl TestService for InteropTestService {
         if req.has_response_status() {
             let code = req.get_response_status().get_code();
             let msg = Some(req.take_response_status().take_message());
-            let status = RpcStatus::new(RpcStatusCode::from(code), msg);
+            let status = RpcStatus::new(RpcStatusCode::new(code), msg);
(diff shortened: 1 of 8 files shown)


src/call/mod.rs
```diff
@@ -31,8 +31,58 @@ use crate::codec::{DeserializeFn, Marshaller, SerializeFn};
...
 use crate::task::{self, BatchFuture, BatchType, CallTag, SpinLock};
 
-pub use crate::grpc_sys::grpc_status_code as RpcStatusCode;
-pub type RpcStatusCodeType = i32;
+/// An gRPC status code structure.
+/// This type contains constants for all gRPC status codes.
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub struct RpcStatusCode(i32);
+
+impl From<i32> for RpcStatusCode {
```
nrc: Do we need this `From` impl? If we don't have it, then we can guarantee that `RpcStatusCode`  only has valid values.

src/call/mod.rs
```diff
@@ -31,8 +31,58 @@ use crate::codec::{DeserializeFn, Marshaller, SerializeFn};
...
+    }
+}
+
+macro_rules! status_codes {
+    (
+        $(
+            ($num:expr, $konst:ident);
+        )+
+    ) => {
+        #[allow(dead_code)]
```
nrc: Why do we need to `allow` here? Since this is all public api it should not be required.

src/call/mod.rs
```diff
@@ -330,9 +380,10 @@ impl BatchContext {
 
     /// Get the status of the rpc call.
     pub fn rpc_status(&self) -> RpcStatus {
-        let status =
-            unsafe { grpc_sys::grpcwrap_batch_context_recv_status_on_client_status(self.ctx) };
-        let details = if status == RpcStatusCode::GRPC_STATUS_OK {
+        let status = RpcStatusCode::from(unsafe {
```
nrc: Are we likely to get values outside the valid range? If not we should `assert` that status is valid, if they might be we might want to at least log invalid values

src/call/mod.rs
BusyJay: I don't think there is such a valid range for code, users are free to use whatever code they want. https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

hunterlxt: convert to `new()` and assert in this function

src/call/mod.rs
hunterlxt: compiler hint, but don't see it this morning. Anyway, solved

src/call/mod.rs
```diff
@@ -31,8 +31,58 @@ use crate::codec::{DeserializeFn, Marshaller, SerializeFn};
...
+        impl RpcStatusCode {
+        $(
+            pub const $konst: RpcStatusCode = RpcStatusCode($num);
+        )+
+        }
+    }
+}
+
+status_codes! {
+    (0, OK);
```
BusyJay: You should use const value from grpc_status_code instead of plain number.

src/call/mod.rs
```diff
@@ -31,8 +31,58 @@ use crate::codec::{DeserializeFn, Marshaller, SerializeFn};
...
-pub use crate::grpc_sys::grpc_status_code as RpcStatusCode;
-pub type RpcStatusCodeType = i32;
+/// An gRPC status code structure.
+/// This type contains constants for all gRPC status codes.
+#[derive(PartialEq, Clone, Copy, Debug)]
+pub struct RpcStatusCode(i32);
+
+impl RpcStatusCode {
+    fn new(code: i32) -> RpcStatusCode {
+        assert!(code >= -1 && code <= 16, "RPC status code is invalid");
```
BusyJay: I think it's wrong to assert here:

> The gRPC client and server-side implementations may also generate and return status on their own when errors happen. Only a subset of the pre-defined status codes are generated by the gRPC libraries. This allows applications to be sure that any other code it sees was actually returned by the application (although it is also possible for the server-side to return one of the codes generated by the gRPC libraries).

https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

hunterlxt: rolled back

src/call/mod.rs
hunterlxt: solved

interop/src/server.rs
```diff
@@ -139,7 +139,7 @@ impl TestService for InteropTestService {
                 if req.has_response_status() {
                     let code = req.get_response_status().get_code();
                     let msg = Some(req.take_response_status().take_message());
-                    let status = RpcStatus::new(RpcStatusCode::new(code), msg);
+                    let status = RpcStatus::new(code.into(), msg);
```
BusyJay: How about changing the args in new to accept `impl Into<RpcStatusCode>`?

src/call/mod.rs
```diff
@@ -397,7 +401,7 @@ impl BatchContext {
             }
         };
 
-        RpcStatus { status, details }
+        RpcStatus::new(status, details)
```
hunterlxt: convert to func new to unify style

interop/src/server.rs
hunterlxt: Done

PR #355, opened 2019-07-25
proto: replace slash to get it built on windows

BusyJay:
rust-protobuf can't detect path correctly, this is a workaround for stepancheg/rust-protobuf#433.

### DIFF
--- a/proto/build.rs
+++ b/proto/build.rs
@@ -29,7 +29,7 @@ fn main() {
                 if !dent.file_type().is_file() {
                     return None;
                 }
-                Some(format!("{}", dent.path().display()))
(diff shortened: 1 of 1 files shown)


proto/build.rs
```diff
@@ -29,7 +29,7 @@ fn main() {
                 if !dent.file_type().is_file() {
                     return None;
                 }
-                Some(format!("{}", dent.path().display()))
+                Some(format!("{}", dent.path().display()).replace('\\', "/"))
```
nrc: Could you add a comment please? This is a very mysterious change unless you know the reason why

PR #356, opened 2019-07-25
grpcio: pre-generate bindings for common targets

hunterlxt:
Building bindings at compile time introduces extra dependencies. And clang is hard to install and configure on some platform like centos 7. It will be very convenient that we can pre-generate bindings for common targets like x86_64-unknown-linux-gnu. For others, we can fallback to rust-bindgen at compile time.


hunterlxt:
ignore the commits before

grpc-sys/src/lib.rs
```diff
@@ -16,7 +16,7 @@
 #![allow(non_upper_case_globals)]
 #[allow(clippy::all)]
 mod bindings {
-    include!(concat!(env!("OUT_DIR"), "/grpc-bindings.rs"));
+    include!(concat!(env!("BINDING_DIR"), "/grpc-bindings.rs"));
```
BusyJay: Better give it a platform-related name. For example, how to name the file if we want to also support macos?

hunterlxt: All Done, PTAL @BusyJay

hunterlxt:
pre-bindings files could be different on different platforms and environments. Even in two computers in the same operating system environment, the order of the files is different. It may be difficult to use `git diff` to verify that the file has been modified. @BusyJay @nrc

hunterlxt:
PLTA @BusyJay

nrc:
Hmm, I agree it is a pain to generate the bindings on every build (especially when using grpc-rs as a dep). However, I'm not sure that this is a good idea: we recently went to a lot of effort to remove generated code for protobufs from the repos, and there seems more risk here than there since the protobuf code is at least multi-platform. It is also not recommended to do this by the bindgen guide (https://rust-lang.github.io/rust-bindgen/library-usage.html).

hunterlxt:
I suggest that use pre-bindings by default, but provide a feature like bindings
-vendored to build your own bindings

.travis.yml
```diff
@@ -100,6 +101,11 @@ before_script:
 script:
   - if [[ $TRAVIS_OS_NAME == "linux" ]] && [[ $TRAVIS_RUST_VERSION == "stable" ]]; then
       rustup component add rustfmt && cargo fmt --all -- --check;
+      scripts/generate-bindings.sh && git diff HEAD;
```
BusyJay: I don't think it will report error if the bindings is modified.

README.md
```diff
@@ -28,9 +28,9 @@ This project is still under development. The following features with the check m
 - Rust >= 1.19.0
 - binutils >= 2.22
 - LLVM and Clang >= 3.9 recommended by bindgen
-- By default, the [secure feature](#feature-secure) is enabled, therefore Go (>=1.7) is required.
+- By default, the [boringssl](#feature-secure) is enabled, therefore Go (>=1.7) is required.
```
BusyJay: I don't think there is such a features named boringssl. Better express in another way.

README.md
```diff
@@ -28,9 +28,9 @@ This project is still under development. The following features with the check m
 - Rust >= 1.19.0
 - binutils >= 2.22
 - LLVM and Clang >= 3.9 recommended by bindgen
```
BusyJay: This is an optional dependency now.

.travis.yml
hunterlxt: I verified，see the log in travis

grpc-sys/build.rs
```diff
@@ -249,40 +250,56 @@ fn bindgen_grpc(mut config: bindgen::Builder) {
...
         .expect("Couldn't write bindings!");
 }
 
 fn main() {
-    let mut cc = Build::new();
-    let mut bind_config = bindgen::Builder::default();
-
     println!("cargo:rerun-if-changed=grpc_wrap.cc");
     println!("cargo:rerun-if-changed=grpc");
+    println!("cargo:rerun-if-changed=bindings");
```
BusyJay: I don't think it's necessary.

README.md
hunterlxt: maybe add （optional）here

grpc-sys/build.rs
```diff
@@ -317,5 +334,43 @@ fn main() {
...
+            );
+        }
+        "x86_64-apple-darwin" => {
+            let file_path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+                .join("bindings")
+                .join("x86_64-apple-darwin-bindings.rs");
+            if env::var("UPDATE_BIND").is_ok() {
+                bindgen_grpc(bind_config, &file_path);
+            }
+            println!(
```
BusyJay: Maybe you can put the print out of the match.

grpc-sys/src/lib.rs
```diff
@@ -16,7 +16,7 @@
 #![allow(non_upper_case_globals)]
 #[allow(clippy::all)]
 mod bindings {
-    include!(concat!(env!("OUT_DIR"), "/grpc-bindings.rs"));
+    include!(concat!(env!("BINDING_DIR")));
```
BusyJay: Better name it BINDING_PATH

README.md
hunterlxt: secure bring misunderstanding, because only boringssl depends on go. maybe change to the following：

the [secure feature](#feature-secure) is enabled and implemented by boringssl

grpc-sys/build.rs
hunterlxt: For example, when you accidentally changed the file here.  You find that everything hasn't changed since you ran the script.  You thought that the binding was generated, but it is not.

.travis.yml
BusyJay: Can you provide a link? As far as I can see, `git diff HEAD` will just display the diff and exit with success code 0.

hunterlxt: Maybe I made a mistake, I will verify and modify it tomorrow.

grpc-sys/build.rs
hunterlxt: you are right

grpc-sys/src/lib.rs
hunterlxt: good suggestion

README.md
BusyJay: How about
```
By default, the secure features is provided by boringssl, which requires Go to build. You can also use openssl instead by enable features openssl.
```
?

@nrc what do you think?

README.md
BusyJay: Better state in what condition they are required explicitly.

grpc-sys/build.rs
BusyJay: Good point! But is it working as expected? As only the modified timestamp of directory bindings instead of actual bindings files is considered.

README.md
nrc: sgtm

nrc:
@BusyJay do you have thoughts on https://github.com/pingcap/grpc-rs/pull/356#issuecomment-516604550 ?

grpc-sys/build.rs
hunterlxt: The timestamp of the file affects the timestamp of the directory, and this is as expected

BusyJay:
> there seems more risk here than there since the protobuf code is at least multi-platform.

That's why the bindings are named according to the target platform. It's intended to only maintain limited platform.

I agree we should remove the pre-generated bindings when rust-bindgen supports bundled clang. For now, pre-generating seems a more suitable solution.

.travis.yml
```diff
@@ -100,6 +101,11 @@ before_script:
 script:
   - if [[ $TRAVIS_OS_NAME == "linux" ]] && [[ $TRAVIS_RUST_VERSION == "stable" ]]; then
       rustup component add rustfmt && cargo fmt --all -- --check;
+      scripts/generate-bindings.sh && git diff --exit-code HEAD;
+    fi
+  - if [[ $TRAVIS_OS_NAME == "osx" ]] && [[ $TRAVIS_RUST_VERSION == "stable" ]]; then
+      rustup component add rustfmt && cargo fmt --all -- --check;
```
BusyJay: Redundant check.

README.md
```diff
@@ -27,10 +27,10 @@ This project is still under development. The following features with the check m
 - CMake >= 3.8.0
 - Rust >= 1.19.0
 - binutils >= 2.22
-- LLVM and Clang >= 3.9 recommended by bindgen
-- By default, the [secure feature](#feature-secure) is enabled, therefore Go (>=1.7) is required.
+- LLVM and Clang >= 3.9 only needed when you want to generate bindings instead of using the default (supported MacOS and Linux)
+- By default, the [secure feature](#feature-secure) is provided by boringssl, which requires Go (>=1.7) to build. You can also use openssl instead by enable [openssl feature](#feature-openssl).
```
BusyJay: s/enable/enabling/

README.md
```diff
@@ -27,10 +27,10 @@ This project is still under development. The following features with the check m
...
 
-For Linux and MacOS, you also need to install gcc (or clang) too.
+For Linux and MacOS, you also need to install gcc too.
```
BusyJay: Why remove clang?

README.md
```diff
@@ -42,7 +42,9 @@ For Windows, you also need to install following software:
 
 ```
 $ git submodule update --init --recursive # if you just cloned the repository
-$ cargo build
+$ cargo build # if you just want to compile
+
```
BusyJay: Redundant blank line

grpc-sys/build.rs
```diff
@@ -249,40 +251,89 @@ fn bindgen_grpc(mut config: bindgen::Builder) {
...
+// need to be updated by default unless the UPDATE_BIND is specified.
+// Other platforms use bindgen to generate the bindings every time.
+fn config_binding_path(config: bindgen::Builder) {
+    let file_path: PathBuf;
+    match env::var("TARGET").unwrap_or("".to_owned()).as_str() {
+        "x86_64-unknown-linux-gnu" => {
+            file_path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+                .join("bindings")
+                .join("x86_64-unknown-linux-gnu-bindings.rs");
+            if env::var("UPDATE_BIND").is_ok() {
```
BusyJay: It may be confusing to still generate bindings when UPDATE_BIND = 0.

grpc-sys/build.rs
BusyJay: I don't think so.
```
/tmp % ls -alh test-dir
total 0
drwxrwxr-x.  2 jay  jay   60 Aug  5 14:56 .
drwxrwxrwt. 22 root root 520 Aug  5 14:56 ..
-rw-rw-r--.  1 jay  jay    0 Aug  5 14:56 testfile
/tmp % touch test-dir/testfile 
/tmp % ls -alh test-dir       
total 0
drwxrwxr-x.  2 jay  jay   60 Aug  5 14:56 .
drwxrwxrwt. 22 root root 520 Aug  5 14:56 ..
-rw-rw-r--.  1 jay  jay    0 Aug  5 14:57 testfile
```

grpc-sys/build.rs
hunterlxt: Sharp eyed

BusyJay:
Because clang is the default compiler on MacOS, so I drop the support. PTAL

grpc-sys/src/lib.rs
```diff
@@ -16,7 +16,7 @@
 #![allow(non_upper_case_globals)]
 #[allow(clippy::all)]
 mod bindings {
-    include!(concat!(env!("OUT_DIR"), "/grpc-bindings.rs"));
+    include!(concat!(env!("BINDING_PATH")));
```
disksing: Can `concat` be removed?

grpc-sys/build.rs
```diff
@@ -249,40 +251,80 @@ fn bindgen_grpc(mut config: bindgen::Builder) {
...
+    println!(
+        "cargo:rustc-env=BINDING_PATH={}",
+        file_path.to_str().unwrap()
+    );
+}
 
+fn main() {
     println!("cargo:rerun-if-changed=grpc_wrap.cc");
     println!("cargo:rerun-if-changed=grpc");
+    println!("rerun-if-env-changed=UPDATE_BIND");
```
Connor1996: no need `cargo:` ?

grpc-sys/build.rs
```diff
@@ -249,40 +251,80 @@ fn bindgen_grpc(mut config: bindgen::Builder) {
...
+        "x86_64-unknown-linux-gnu" => {
+            file_path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+                .join("bindings")
+                .join("x86_64-unknown-linux-gnu-bindings.rs");
+            if env::var("UPDATE_BIND").map(|s| s == "1").unwrap_or(false) {
+                bindgen_grpc(config, &file_path);
+            }
+        }
+        _ => {
+            file_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("grpc-bindings.rs");
```
Connor1996: I am confusing at why the file path is not consistent with above

BusyJay: Because for x86_64 linux, bindings are pre-generated, so that clang is not required any more. That means the generated bindings will be shipped with all the other source code. See more in the description of the PR.

.travis.yml
hunterlxt: @BusyJay So how do you know if the person using mac has run this script?

hunterlxt: ignore this, I didn't see you killing the mac

PR #363, opened 2019-08-26
unsafe impl Send for Metadata

### LINKED ISSUE #362: unsafe impl `Send` for `Metadata`
I would like to share a `grpcio::CallOption` between threads. This isn't (safely) possible at the moment because the raw pointers in it prevent Rust from deriving `Send`.

As best I can tell, `grpcio::Metadata` owns all of its data, so it can explicitly be given `unsafe impl Send`. This would allow Rust to auto-derive `Send` for `grpcio::CallOption`.

Manually giving `unsafe impl Send` to types that I use that contain a `grpcio::CallOption`... Or making my own struct that gets converted into a `grpcio::CallOption` on the fly.

Spferical:
Fixes #362

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -218,6 +218,8 @@ impl Drop for Metadata {
     }
 }
 
+unsafe impl Send for Metadata {}
+
 /// Immutable metadata iterator
 ///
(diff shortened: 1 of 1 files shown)


Spferical:
(rebased onto master)

PR #374, opened 2019-09-11
Update grpc to 1.22.1

nrc:
This pulls the submodule from grpc/grpc instead of PingCAP/grpc, since the only extra we seem to have is a commit for benchmarking Rust. Therefore, it seems better for us not to use a fork.

This passes all the tests, but it seems too easy...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.5.0-alpha.4"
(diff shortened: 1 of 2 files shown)


nrc:
hmm, seems I need to regenerate scripts and for that I need a Linux box :-( Guess I'll spin up a cloud machine tomorrow. Closing PR for now.

PR #377, opened 2019-09-18
grpcio: add resource quota support

hunterlxt:
* Add the `ResourceQuota` struct to control the bound on memory usage.
* Wrap `set_pointer_vtable` to configure channel args.
* Format `grpc_wrap.cc` code style.

### DIFF
--- a/grpc-sys/grpc_wrap.cc
+++ b/grpc-sys/grpc_wrap.cc
@@ -105,7 +105,7 @@ typedef struct grpcwrap_batch_context {
 GPR_EXPORT grpcwrap_batch_context* GPR_CALLTYPE
 grpcwrap_batch_context_create() {
   auto* ctx =
-    (grpcwrap_batch_context*)gpr_malloc(sizeof(grpcwrap_batch_context));
+      (grpcwrap_batch_context*)gpr_malloc(sizeof(grpcwrap_batch_context));
   memset(ctx, 0, sizeof(grpcwrap_batch_context));
   return ctx;
 }
@@ -118,8 +118,8 @@ typedef struct {
 
 GPR_EXPORT grpcwrap_request_call_context* GPR_CALLTYPE
 grpcwrap_request_call_context_create() {
-  auto * ctx = (grpcwrap_request_call_context*)
-    gpr_malloc(sizeof(grpcwrap_request_call_context));
+  auto* ctx = (grpcwrap_request_call_context*)gpr_malloc(
+      sizeof(grpcwrap_request_call_context));
   memset(ctx, 0, sizeof(grpcwrap_request_call_context));
   return ctx;
 }
@@ -295,35 +295,32 @@ grpcwrap_batch_context_recv_initial_metadata(
   return &(ctx->recv_initial_metadata);
 }
 
-GPR_EXPORT const char *GPR_CALLTYPE
-grpcwrap_slice_raw_offset(const grpc_slice *slice, size_t offset, size_t *len) {
+GPR_EXPORT const char* GPR_CALLTYPE
+grpcwrap_slice_raw_offset(const grpc_slice* slice, size_t offset, size_t* len) {
   *len = GRPC_SLICE_LENGTH(*slice) - offset;
-  return (const char *)(GRPC_SLICE_START_PTR(*slice)) + offset;
+  return (const char*)(GRPC_SLICE_START_PTR(*slice)) + offset;
 }
 
 GPR_EXPORT grpc_slice GPR_CALLTYPE
-grpcwrap_slice_copy(const grpc_slice *slice) {
+grpcwrap_slice_copy(const grpc_slice* slice) {
   return grpc_slice_copy(*slice);
 }
 
-GPR_EXPORT void GPR_CALLTYPE
-grpcwrap_slice_unref(const grpc_slice *slice) {
+GPR_EXPORT void GPR_CALLTYPE grpcwrap_slice_unref(const grpc_slice* slice) {
   grpc_slice_unref(*slice);
 }
 
-GPR_EXPORT grpc_slice GPR_CALLTYPE
-grpcwrap_slice_ref(const grpc_slice *slice) {
+GPR_EXPORT grpc_slice GPR_CALLTYPE grpcwrap_slice_ref(const grpc_slice* slice) {
   return grpc_slice_ref(*slice);
 }
 
-GPR_EXPORT size_t GPR_CALLTYPE
-grpcwrap_slice_length(const grpc_slice *slice) {
+GPR_EXPORT size_t GPR_CALLTYPE grpcwrap_slice_length(const grpc_slice* slice) {
   return GRPC_SLICE_LENGTH(*slice);
 }
 
-GPR_EXPORT grpc_byte_buffer *GPR_CALLTYPE
-grpcwrap_batch_context_take_recv_message(grpcwrap_batch_context *ctx) {
-  grpc_byte_buffer *buf = nullptr;
+GPR_EXPORT grpc_byte_buffer* GPR_CALLTYPE
+grpcwrap_batch_context_take_recv_message(grpcwrap_batch_context* ctx) {
+  grpc_byte_buffer* buf = nullptr;
   if (ctx->recv_message) {
     buf = ctx->recv_message;
     ctx->recv_message = nullptr;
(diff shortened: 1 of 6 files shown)


hunterlxt:
@BusyJay  PTAL

src/quota.rs
```diff
@@ -0,0 +1,29 @@
+use crate::grpc_sys::{self, grpc_resource_quota};
+use std::ptr;
+
+/// ResourceQuota represents a bound on memory and thread usage by the gRPC,
+/// Now grpc-rs only supports setting memory bound. A ResourceQuota can be
+/// consumed by a ChannelBuilder.
+pub struct ResourceQuota {
+    pub raw: *mut grpc_resource_quota,
```
BusyJay: Should not be a public field.

src/channel.rs
```diff
@@ -85,6 +87,7 @@ fn dur_to_ms(dur: Duration) -> i32 {
 enum Options {
     Integer(i32),
     String(CString),
+    Pointer(*mut c_void, *mut c_void),
```
BusyJay: Memory is leak if options is not used.

src/quota.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+impl ResourceQuota {
+    /// Create a control block for resource quota
+    pub fn new(name: Option<String>) -> ResourceQuota {
```
BusyJay: Seems `Option<&str>` is enough. By the way, is empty string a valid argument?

src/channel.rs
hunterlxt: Can you explain in detail?

src/quota.rs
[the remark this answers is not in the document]
hunterlxt: Now it's changed to reference. Is memory leak ?

hunterlxt: Yeah, and core will handle the valid string.

src/quota.rs
hunterlxt: Add a public function to solve

src/channel.rs
BusyJay: For example, if the builder is created and configured, but never call `build_args` or `connect`, and then dropped.

src/quota.rs
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    /// Resize this ResourceQuota to a new memory size.
+    pub fn resize(self, new_size: usize) -> ResourceQuota {
+        unsafe { grpc_sys::grpc_resource_quota_resize(self.raw, new_size) };
+        self
+    }
+
+    pub fn get_raw(&self) -> *mut grpc_resource_quota {
```
BusyJay: Better be `pub(crate)`.

src/quota.rs
```diff
@@ -0,0 +1,42 @@
...
+        self.raw
+    }
+}
+
+impl Drop for ResourceQuota {
+    fn drop(&mut self) {
+        println!("drop ResourceQuota");
```
BusyJay: Why println?

grpc-sys/grpc_wrap.cc
```diff
@@ -450,6 +450,16 @@ GPR_EXPORT void GPR_CALLTYPE grpcwrap_channel_args_set_integer(
   args->args[index].value.integer = value;
 }
 
+GPR_EXPORT void GPR_CALLTYPE grpcwrap_channel_args_set_pointer_vtable(
+    grpc_channel_args* args, size_t index, const char* key, void* value, const grpc_arg_pointer_vtable* vtable) {
+  GPR_ASSERT(args);
+  GPR_ASSERT(index < args->num_args);
+  args->args[index].type = GRPC_ARG_POINTER;
+  args->args[index].key = gpr_strdup(key);
+  args->args[index].value.pointer.p = vtable->copy(value);
```
BusyJay: Can it be moved?

src/channel.rs
nrc: I think it might be good to have a struct called something like PointerVtable with named fields (having two `*mut c_void` pointers scares me) and then you can call the destroy function in the destructor.

src/quota.rs
```diff
@@ -0,0 +1,42 @@
...
+pub struct ResourceQuota {
+    raw: *mut grpc_resource_quota,
+}
+
+impl ResourceQuota {
+    /// Create a control block for resource quota
+    pub fn new(name: Option<&str>) -> ResourceQuota {
+        match name {
+            Some(name_str) => ResourceQuota {
+                raw: unsafe { grpc_sys::grpc_resource_quota_create(name_str.as_ptr() as _) },
```
nrc: I don't think this is safe - the name may not live long enough. I think it is best to only take `&'static str`

nrc: And I think the string should be null-terminated, so either take a C str or make it unsafe and add a comment

src/quota.rs
hunterlxt: Forgot to remove test code😂

src/quota.rs
BusyJay: The method will clone the string. But null-terminated is a good point! 👍

src/channel.rs
hunterlxt: @BusyJay @nrc 
Done! `ChannelBuilder` is responsible for destroying the pointer of its own reference when destroying.

src/quota.rs
```diff
@@ -0,0 +1,45 @@
+use crate::grpc_sys::{self, grpc_resource_quota};
+use std::ffi::CString;
+use std::ptr;
+
+/// ResourceQuota represents a bound on memory and thread usage by the gRPC
```
BusyJay: Better comment why thread quota doesn't work at the moment.

src/quota.rs
```diff
@@ -0,0 +1,45 @@
...
+                }
+            }
+            None => ResourceQuota {
+                raw: unsafe { grpc_sys::grpc_resource_quota_create(ptr::null()) },
+            },
+        }
+    }
+
+    /// Resize this ResourceQuota to a new memory size.
+    pub fn resize(self, new_size: usize) -> ResourceQuota {
```
BusyJay: It's confusing. How about `resize_memory`?

src/quota.rs
```diff
@@ -0,0 +1,45 @@
...
+        unsafe { grpc_sys::grpc_resource_quota_resize(self.raw, new_size) };
+        self
+    }
+
+    pub(crate) fn get_ptr(&self) -> *mut grpc_resource_quota {
```
BusyJay: It seems what you need is `into_ptr(self) -> *mut grpc_resource_quota`, so that additional ref can be omitted.

src/channel.rs
```diff
@@ -129,6 +142,22 @@ impl ChannelBuilder {
...
 
+    /// Set resource quota by consuming a ResourceQuota
+    pub fn set_resource_quota(mut self, quota: ResourceQuota) -> ChannelBuilder {
+        let raw_quota = quota.get_ptr();
+        unsafe {
+            grpc_sys::grpc_resource_quota_ref(raw_quota);
+            self.options.insert(
+                Cow::Borrowed(grpc_sys::GRPC_ARG_RESOURCE_QUOTA),
+                Options::Pointer(
+                    raw_quota as _,
```
BusyJay: Why not store it as type `ResourceQuota`? So that explicit `Drop` impl is not necessary.

grpc-sys/grpc_wrap.cc
BusyJay: So that create a quota and configure it requires 2 ref operations. And using move mechanism these two refs can be avoided.

src/channel.rs
hunterlxt: Good Point, `ResourceQuota` don't have to deal with destructuring

grpc-sys/grpc_wrap.cc
hunterlxt: @BusyJay  Now remove a ref in `ChannelBuilder::set_resource_quota`

src/channel.rs
```diff
@@ -85,6 +87,7 @@ fn dur_to_ms(dur: Duration) -> i32 {
 enum Options {
     Integer(i32),
     String(CString),
+    Pointer(ResourceQuota, *mut c_void),
```
BusyJay: Why not use its concrete type?

hunterlxt: `*const grpc_arg_pointer_vtable` ?

BusyJay: Yes.

src/channel.rs
```diff
@@ -453,13 +454,13 @@ impl ChannelBuilder {
...
                 },
-                Options::Pointer(ref quota, p2) => unsafe {
+                Options::Pointer(ref quota, vtable) => unsafe {
                     grpc_sys::grpcwrap_channel_args_set_pointer_vtable(
                         args,
                         i,
                         key,
                         quota.get_ptr() as _,
-                        p2 as _,
+                        vtable as _,
```
BusyJay: Unnecessary cast.

hunterlxt: Done!

PR #388, opened 2019-10-22
Benchmark: add control driver

hunterlxt:
The control side of the benchmark is provided [here](https://github.com/pingcap/grpc/tree/4158bd07777a147f95f9df03efdfe924be43a956/test/cpp/qps), although it can be used, but it is very troublesome. Now we need to provide some custom tests, such as stability related, so we need our own driver.

**Note**: The current implementation is not perfect, because this PR is too big, I need to split it into two, the first step to complete the driver, the second step to enrich the tikv related scenario.

[issue](https://github.com/pingcap/grpc-rs/issues/385)


hunterlxt:
This PR is staled. Closed it.

Issue #400, opened 2019-11-11
Too many allocations in grpcio


BusyJay:
For steaming calls, every subsequent messages will allocate a Batch OpSet and a new tag. By design grpc core does not expect user to send next message for the same call until last tag is returned. Hence these two allocations can be reused messages.

To reuse batch and tag, we need to refactor current `CallTag` implement to allow a long live batch. batch should also be implemented in rust as operation sets so that we have more flexible and safe way to refactor.

BusyJay:
I build a POC in recent days, and early benches show that it has at most 18% throughput improvement.

PR #401, opened 2019-11-11
util/buf: rusty grpc_slice and its related structs

BusyJay:
This PR wraps `grpc_slice` into a safe struct and provides convenient
and safe APIs to manipulate and transform. `grpc_slice` is now `Copy`
and we should use its wrapper `GrpcSlice` to make it clone correctly.

This is the first step to reduce allocations in grpcio, see also #400.

This PR has a lot of common changes with #382. The most notable
difference is the implement of vtable struct.


src/util/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod buf;
```
nrc: Rather than add a util module, could we make buf.rs a top level module? I also prefer buffers to buf, but that is not too important.

src/util/buf.rs
```diff
@@ -0,0 +1,560 @@
...
+
+/// A convenient rust wrapper for the type `grpc_slice`.
+#[repr(C)]
+pub struct GrpcSlice(grpc_slice);
+
+impl GrpcSlice {
+    /// Allocate a `GrpcSlice` with given length.
+    ///
+    /// It's unsafe to read from the slice as the content is uninitialized.
+    pub unsafe fn with_len(len: usize) -> GrpcSlice {
```
nrc: Is this method used? I don't see it anywhere. Could it be removed to shrink the unsafe surface area?

src/util/buf.rs
```diff
@@ -0,0 +1,560 @@
...
+                let len = self.0.data.refcounted.length;
+                std::slice::from_raw_parts(start, len)
+            } else {
+                let len = self.0.data.inlined.length;
+                &self.0.data.inlined.bytes[..len as usize]
+            }
+        }
+    }
+
+    pub fn as_ptr(&self) -> *const u8 {
```
nrc: This also seems to be unused and makes it much harder to ensure Rust's immutability invariant. Unless you have a use in mind, I would remove it.

src/util/buf.rs
```diff
@@ -0,0 +1,560 @@
...
+use std::ffi::c_void;
+use std::fmt::{self, Debug, Formatter};
+use std::io::{self, BufRead, Read};
+use std::mem::{self, ManuallyDrop, MaybeUninit};
+use std::ptr;
+use std::sync::atomic::{self, AtomicUsize, Ordering};
+
+/// A convenient rust wrapper for the type `grpc_slice`.
+#[repr(C)]
+pub struct GrpcSlice(grpc_slice);
```
nrc: Can the user assume that the inner grpc_slice is always valid?

src/util/buf.rs
```diff
@@ -0,0 +1,560 @@
...
+/// When ref count decreases to 0, there will be no access to the data.
+/// To utilize the mechanism, we need to define a struct which has
+/// `grpc_slice_refcount` type as the first field, so that it's safe to cast
+/// a pointer to the struct to `grpc_slice_refcount` and all ref/unref operations
+/// will be forwarded to our own implement.
+///
+/// Vec will be stored inside the struct so that it will be dropped automatically
+/// once the struct is dropped.
+#[repr(C)]
+struct VecRefCount {
```
nrc: Could you document that `VecRefCount` must not be moved

nrc: You could also add a `PhantomPinned` field so the compiler can help with enforcing that.

src/util/buf.rs
```diff
@@ -0,0 +1,560 @@
...
+    ///
+    /// A buffer, which length is 1, is allocated for the slice.
+    #[allow(clippy::cast_ref_to_mut)]
+    fn from(s: &'a GrpcSlice) -> GrpcByteBuffer {
+        unsafe {
+            // hack: buffer_create accepts an mutable pointer to indicate it mutate
+            // ref count. Ref count is recorded by atomic variable, which is considered
+            // `Sync` in rust. This is an interesting difference in what is *mutable*
+            // between C++ and rust.
+            let s = &mut *(s as *const GrpcSlice as *mut GrpcSlice);
```
nrc: This is UB; you should an `UnsafeCell` here in order to get the `*mut` pointer

src/util/buf.rs
```diff
@@ -0,0 +1,560 @@
...
+
+/// A zero-copy reader for the message payload.
+///
+/// To achieve zero-copy, use the BufRead API `fill_buf` and `consume`
+/// to operate the reader.
+#[repr(C)]
+pub struct GrpcByteBufferReader {
+    reader: grpc_byte_buffer_reader,
+    /// Current reading buffer.
+    slice: ManuallyDrop<GrpcSlice>,
```
nrc: Please document why this is `ManuallyDrop`

src/util/buf.rs
BusyJay: Yes.

src/util/buf.rs
BusyJay: Why? It can be moved around.

src/util/buf.rs
BusyJay: `UnsafeCell` will make the implement very verbose. This is the only place that does the hack, and it's put inside unsafe already.

src/util/buf.rs
nrc: Great! Could you add it to the docs please?

src/util/buf.rs
nrc: It does unfortunately make it more verbose but I think that casting like this is UB, whether it is safe or not because it can break LLVM's alias analysis. Adding the `UnsafeCell` means rustc will not tell LLVM that `s` is unique (I'm not 100% sure, I know this is true for `&` pointers, but I can't find documentation about `*` pointers, however I think the compiler treats them the same and it is better to be safe, I think).

grpc-sys/build.rs
```diff
@@ -309,18 +309,23 @@ fn bindgen_grpc(mut config: bindgen::Builder, file_path: &PathBuf) {
...
         .blacklist_type(r"gpr_cv")
         .blacklist_type(r"gpr_once")
         .constified_enum_module(r"grpc_status_code")
-        .no_copy("grpc_slice")
         .default_enum_style(bindgen::EnumVariation::Rust {
             non_exhaustive: false,
-        })
-        .generate()
+        });
+    println!("running {}", cfg.command_line_flags().join(" "));
```
overvenus: Is this necessary? I think `cargo build --vv` should output something similar.

grpc-sys/bindings/x86_64-unknown-linux-gnu-bindings.rs
```diff
@@ -706,6 +672,7 @@ fn bindgen_test_layout_grpc_slice() {
 #[doc = " Represents an expandable array of slices, to be interpreted as a"]
 #[doc = "single item."]
 #[repr(C)]
+#[derive(Copy, Clone)]
 pub struct grpc_slice_buffer {
```
overvenus: Is it ok to copy the slice? How do we manage the ownership of allocated memory?

BusyJay: Yes, it's a bit-wise copy of the struct value, so users have to deal with ownership carefully. I think it's OK for the C type to be copy and clone. For example, `libc::iovec` is also `Copy` and `Clone`, but calling clone will not actually clone the inner data. Make it `Copy` is handy and necessary for fighting with FFIs.

grpc-sys/build.rs
BusyJay: Tried and couldn't find it from stdout, stderr, $OUT_DIR/../output, $OUT_DIR/../stderr.

src/buf.rs
```diff
@@ -0,0 +1,637 @@
+// Copyright 2019 PingCAP, Inc.
```
hicqu: Should be
```
// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```

src/buf.rs
```diff
@@ -0,0 +1,637 @@
...
+            if 0 == grpc_byte_buffer_reader_next(reader.as_mut_ptr(), s.as_mut_ptr()) {
+                s.as_mut_ptr().write(grpc_empty_slice());
+            }
+            let remain = grpc_byte_buffer_length((*reader.as_mut_ptr()).buffer_out);
+            // buf is stored inside `reader` as `buffer_in`, so do not drop it.
+            mem::forget(buf);
+
+            GrpcByteBufferReader {
+                reader: reader.assume_init(),
+                slice: ManuallyDrop::new(GrpcSlice(s.assume_init())),
```
hicqu: Could `grpc_empty_slice()` be uninitialized?

hicqu: Better to add a test case for this.

BusyJay: The purpose of `grpc_empty_slice` is to create an empty slice, so it has to be initialized. It's guaranteed by the grpc c core.

hicqu: Got it.

BusyJay:
grpc/grpc#18407 revoke the access to grpc_slice_refcount, this PR can't work with latest release of grpc. /cc @hicqu  please check if it affects your work.

Issue #399, opened 2019-11-10
Why are documents of channel_args and build_args hidden?


tony612:
**Is your feature request related to a problem? Please describe.**

I found documents of `ServerBuilder::channel_args` and `ChannelBuilder::build_args` are hidden. And comments of `build_args` says `users should use the encapsulated API instead.`, but I can't find the "encapsulated API". So how to set options on server side?

For example, gRPC C++'s [ServerBuilder](https://grpc.github.io/grpc/cpp/classgrpc__impl_1_1_server_builder.html) has several functions `Set*` to set the options, but I didn't find how to do this except using `ServerBuilder::channel_args`.

**Describe the solution you'd like**

I think we'd better make docs of these two functions public or/and we can add functions in `ServerBuilder` to set the options.

BusyJay:
You are right, they should be public APIs. Can you help us to make them public (and probably with better comments)?

tony612:
OK. I can help with this.

PR #405, opened 2019-11-15
doc: make build_args and channel_args public

tony612:
Fixes https://github.com/tikv/grpc-rs/issues/399


tony612:
@BusyJay It seems the CI failure is not caused by my changes. Could you verify? Thank you!

BusyJay:
Your change looks good. /cc @overvenus PTAL

PR #406, opened 2019-11-21
Support openssl-vendored feature

overvenus:
Cherry-pick #262 #326


BusyJay:
Then you also need to pick #387.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.4.7 - 2019-11-20
```
BusyJay: Should not add a new version as CI problem is not fixed yet.

overvenus:
PTAL, thanks!

PR #410, opened 2019-11-26
build script: adjust build for ppcle64

BusyJay:
Allow whitelist recursively to get around platform related issues.


grpc-sys/bindings/x86_64-unknown-linux-gnu-bindings.rs
```diff
@@ -762,6 +772,11 @@ fn bindgen_test_layout_grpc_slice_buffer() {
         )
     );
 }
+impl ::std::fmt::Debug for grpc_slice_buffer {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
+        write ! ( f , "grpc_slice_buffer {{ base_slices: {:?}, slices: {:?}, count: {:?}, capacity: {:?}, length: {:?}, inlined: {:?} }}" , self . base_slices , self . slices , self . count , self . capacity , self . length , self . inlined )
```
overvenus: The format looks weird.

grpc-sys/build.rs
```diff
@@ -288,7 +288,8 @@ fn bindgen_grpc(mut config: bindgen::Builder, file_path: &PathBuf) {
         .clang_arg("-xc++")
         .clang_arg("-I./grpc/include")
         .clang_arg("-std=c++11")
-        .whitelist_recursively(false)
```
overvenus: What does it resolve?

BusyJay: If `recusive` is false, all basic types have to be whitelisted explicitly due to limitation of rust-lang/rust-bindgen#1454.

PR #413, opened 2019-11-28
bump to 0.4.7

hicqu:


hicqu:
PTAL @BusyJay @overvenus thanks! We need to use it with tikv 3.0.6 later.

BusyJay:
Should wait #411.

PR #419, opened 2019-12-20
*: use libz-sys instead of bundled one

BusyJay:
When using the built-in cmake to build zlib, it changes the source tree
as madler/zlib#162 describes. This leads to the failure during
[generating the docs][1]. So let's switch to libz-sys instead, which
uses its own custom script to build zlib, and leave source tree as it
is. Switching to libz-sys can also reduce the package size as we can
ignore more sub modules. It should improve compile time if libz-sys is
also a dependency of other crates.

The only shortcoming is that libz-sys may not be compatible with
grpcio, but I believe the chance is quite small given it's such a small
library. And giving it's such a small library, the benifits like compile
time or package size described above may be too small to be observed.

[1]: https://docs.rs/crate/grpcio/0.5.0-alpha.5/builds/196235.

### DIFF
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -230,14 +230,16 @@ fn setup_libz(config: &mut Config) {
     config.register_dep("Z");
     // cmake script expect libz.a being under ${DEP_Z_ROOT}/lib, but libz-sys crate put it
     // under ${DEP_Z_ROOT}/build. Append the path to CMAKE_PREFIX_PATH to get around it.
-    env::set_var("CMAKE_PREFIX_PATH", {
-        let zlib_path = format!("{}/build", env::var("DEP_Z_ROOT").unwrap());
-        if let Ok(prefix_path) = env::var("CMAKE_PREFIX_PATH") {
-            format!("{};{}", prefix_path, zlib_path)
-        } else {
-            zlib_path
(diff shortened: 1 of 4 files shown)


grpc-sys/build.rs
```diff
@@ -231,6 +225,21 @@ fn figure_ssl_path(build_dir: &str) {
...
 
+fn setup_libz(config: &mut Config) {
+    config.define("gRPC_ZLIB_PROVIDER", "package");
+    config.register_dep("Z");
+    // cmake script expect libz.a being under ${DEP_Z_ROOT}/lib, but libz-sys crate put it
+    // under ${DEP_Z_ROOT}/build. Append the path to CMAKE_PREFIX_PATH to get around it.
+    env::set_var("CMAKE_PREFIX_PATH", {
+        let zlib_path = format!("{}/build", env::var("DEP_Z_ROOT").unwrap());
+        if let Ok(prefix_path) = env::var("CMAKE_PREFIX_PATH") {
+            format!("{};{}", prefix_path, zlib_path)
```
hicqu: Bad format string?

BusyJay: Why?

BusyJay:
@hicqu @nrc PTAL

PR #424, opened 2020-01-08
bump to 0.3.2

hicqu:


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.3.2 - 2020-01-08
+- fix a potential memory leak bug
```
BusyJay: I don't think it's leak.

hicqu: PTAL

PR #425, opened 2020-01-08
*: update to grpc 1.26.0

BusyJay:
Close #375.


src/buf.rs
```diff
@@ -107,84 +105,33 @@ impl PartialEq<GrpcSlice> for GrpcSlice {
...
-                    sub_refcount: ptr::null_mut(),
-                },
-                refs: AtomicUsize::new(1),
-                v,
-            });
-            ref_count.rc.sub_refcount = &mut ref_count.rc;
-            slice.0.refcount = Box::into_raw(mem::transmute(ref_count));
-            slice
+        unsafe {
+            GrpcSlice(grpcio_sys::grpc_slice_from_copied_buffer(
```
hicqu:
```cpp
grpc_slice grpc_slice_new_with_len(void* p, size_t len,
                                   void (*destroy)(void*, size_t)) {
  grpc_slice slice;
  slice.refcount = (new grpc_core::NewWithLenSliceRefcount(destroy, p, len))
                       ->base_refcount();
  slice.data.refcounted.bytes = static_cast<uint8_t*>(p);
  slice.data.refcounted.length = len;
  return slice;
}
```
Seems we can pass `capacity` into `grpc_slice_new_with_len`, and then
```c
slice.data.refcounted.length = v.len();
```

BusyJay: I prefer not to rely on the implement details of `grpc_slice_new_with_len`.

hicqu: I think just not to rely `grpc_core::NewWithLenSliceRefcount` is ok. `slice.data.refcounted.length` is public for us, so we can adjust it if only we can ensure it's less than `capacity`.
Otherwise memory copy is necessary, I think we need to avoid it.

BusyJay: If not rely on `grpc_core::NewWithLenSliceRefCount`, then you can't for sure whether `slice.data.refcounted.length` or its internal field is forwarded to destroy hook.

BusyJay:
@nrc @hicqu PTAL.

One breakage is that grpcio probably can't link statically on MacOS that is <= 10.13. A similar issue can be found alexcrichton/curl-rust#279.

nrc:
> One breakage is that grpcio probably can't link statically on MacOS that is <= 10.13

Seems OK, I think most Mac devs stay up to date.

PR #427, opened 2020-01-15
*: rearrange ssl related code

BusyJay:
To keep code clean and prepare to add authentication support.


src/security/credentials.rs
```diff
@@ -171,8 +231,27 @@ impl ChannelCredentialsBuilder {
...
         );
 
-        let creds =
-            unsafe { grpc_sys::grpcwrap_ssl_credentials_create(root_ptr, cert_ptr, key_ptr) };
+        let mut pair = grpcio_sys::grpc_ssl_pem_key_cert_pair {
+            private_key: key_ptr,
+            cert_chain: cert_ptr,
+        };
+        let creds = unsafe {
+            if cert_ptr.is_null() {
```
hunterlxt: Can you explain why want to add this judgment? I think if the user does not input correctly, it should panic instead of not passing in null by default.

BusyJay: It's a valid usage allowed by grpc.

hunterlxt:
@hicqu PTAL

src/security/credentials.rs
```diff
@@ -5,7 +5,71 @@ use std::ptr;
...
+
+#[repr(u32)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum CertificateRequestType {
+    /// Server does not request client certificate.
+    ///
+    /// The certificate presented by the client is not checked by the server at
+    /// all. (A client may present a self signed or signed certificate or not
+    /// present a certificate at all and any of those option would be accepted)
+    DontRequestClientCertificate = 0,
```
overvenus:
```suggestion
    DontRequestClientCertificate = grpcio_sys::grpc_ssl_client_certificate_request_type::GRPC_SSL_DONT_REQUEST_CLIENT_CERTIFICATE,
```

How about using `grpcio_sys::grpc_ssl_client_certificate_request_type::GRPC_SSL_DONT_REQUEST_CLIENT_CERTIFICATE`?

src/security/credentials.rs
overvenus: What about returning an error instead?

BusyJay: It's a valid usage that should not report error or panic. The statement "user does not input correctly" is just wrong.

hunterlxt: Yeah, just root cert is OK, but incomplete cert_chain is not ok.

hunterlxt:
error: Constants have by default a `'static` lifetime
  --> grpc-sys/build.rs:18:22
   |
18 | const GRPC_VERSION: &'static str = "1.26.0";
   |                     -^^^^^^^---- help: consider removing `'static`: `&str`
   |
   = note: `-D clippy::redundant-static-lifetimes` implied by `-D clippy::all`

PR #428, opened 2020-01-21
Remove deprecated usage of Error trait

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>


BusyJay:
@overvenus @hicqu PTAL

PR #433, opened 2020-02-26
Better Display impl for grpcio::Error

### LINKED ISSUE #432: Better Display impl for `grpcio::Error`
The current [Display implementation](https://docs.rs/grpcio/0.5.0-alpha.4/src/grpcio/error.rs.html#27-48) for `grpcio::Error` is:
```rust
impl Display for Error {
    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
        write!(fmt, "{:?}", self)
    }
}
```
Unfortunatly this doesn't output a very user friendly error in the case of a RpcFailure `RpcFailure(RpcStatus { status: RpcStatusCode(13), details: Some("") })`, notably it forces the developer to know the mapping code to name for each RpcStatusCode. 

Implement `Display` on `RpcStatus` and `RpcStatusCode` and update the impl for `Display` for grpcio::Error to:
```rust
impl Display for Error {
    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
        match self {
            Error::RpcFailure(rpc_failure) => write!(fmt, "RpcFailure({})", rpc_failure),
            other_error => write!(fmt, "{:?}", other_error),
       }
    }
}
```

If you validate this solution I'll gladly submit a PR.

**Additional context**
Have a nice day !

Elrendio:
Resolves #432

**What it does:**
 - Impl Display for RpcStatus and RpcStatusCode
 - Update Display impl for grpcio::Error

Signed-off-by: Elrendio <oscar.walter@ens.fr>

### DIFF
--- a/src/call/mod.rs
+++ b/src/call/mod.rs
@@ -4,7 +4,10 @@ pub mod client;
 pub mod server;
 
 use std::sync::Arc;
-use std::{ptr, slice};
+use std::{
+    fmt::{self, Display, Formatter},
+    ptr, slice,
+};
 
 use crate::cq::CompletionQueue;
 use crate::grpc_sys::{self, grpc_call, grpc_call_error, grpcwrap_batch_context};
@@ -40,13 +43,26 @@ impl Into<i32> for RpcStatusCode {
 macro_rules! status_codes {
     (
         $(
-            ($num:expr, $konst:ident);
+            ($num:path, $konst:ident);
         )+
     ) => {
         impl RpcStatusCode {
         $(
             pub const $konst: RpcStatusCode = RpcStatusCode($num);
         )+
+        }
+        impl Display for RpcStatusCode {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                write!(
+                    f,
+                    "{}-{}",
+                    self.0,
+                    match self {
+                        $(RpcStatusCode($num) => stringify!($konst),)+
+                        RpcStatusCode(_) => "INVALID_STATUS_CODE",
+                    }
+                )
+            }
         }
     }
 }
@@ -138,6 +154,16 @@ pub struct RpcStatus {
 
     /// Optional detail string.
     pub details: Option<String>,
+}
+
+impl Display for RpcStatus {
+    fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            fmt,
+            "RpcStatus {{ status: {}, details: {:?} }}",
+            self.status, self.details
+        )
+    }
 }
 
 impl RpcStatus {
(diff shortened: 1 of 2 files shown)


Elrendio:
The tests don't pass but it really doesn't seem related to my PR. Notably the clippy warning concerns a completely different block of code...

src/call/mod.rs
```diff
@@ -25,6 +28,36 @@ const BUF_SHRINK_SIZE: usize = 4 * 1024;
...
+                8 => "RESOURCE_EXHAUSTED",
+                9 => "FAILED_PRECONDITION",
+                10 => "ABORTED",
+                11 => "OUT_OF_RANGE",
+                12 => "UNIMPLEMENTED",
+                13 => "INTERNAL",
+                14 => "UNAVAILABLE",
+                15 => "DATA_LOSS",
+                16 => "UNAUTHENTICATED",
+                _ => "INVALID_STATUS_CODE",
```
BusyJay: How about adding the number to the message?

src/error.rs
```diff
@@ -38,7 +38,10 @@ pub enum Error {
 
 impl Display for Error {
     fn fmt(&self, fmt: &mut Formatter<'_>) -> fmt::Result {
-        write!(fmt, "{:?}", self)
+        match self {
+            Error::RpcFailure(rpc_failure) => write!(fmt, "RpcFailure({})", rpc_failure),
```
BusyJay: "RpcFailure(RpcStatus {...})" seems a little verbose to me. How about just keeping "RpcFailure"?

src/call/mod.rs
Elrendio: The number is always added to the message (see line 35 and second arg of format), so the messages will be `13-INTERNAL` or `35-INVALID_STATUS_CODE`

src/error.rs
Elrendio: So something like:
`RpcFailure({ status: 13-INTERVAL, details: Some("Something very bad happened") })` ?

I'm not sure just `RpcFailure` is enough because it really nice to have the status in the error message.

I don't really like: `RpcFailure { status: 13-INTERNVAL, details: Some("Something very bad happened") }` because it makes RpcFailure look like a struct when it's a enum. We could also change the RpcFailure variant definition to fit the error:
**From:**
```rust
enum Error {
    RpcFailure(RpcStatus)
}
```

**To** 
```rust
enum Error {
    RpcFailure {
       status: RpcStatusCode,
       code: Option<String>
    }
}
```

But that would be a breaking change.

BusyJay:
@hunterlxt PTAL

BusyJay:
It seemed you didn't sign off your commits correctly.

Elrendio:
Squashed and rebased my commits to have one signature

PR #437, opened 2020-02-29
Improve `Debug` impl of `RpcStatusCode`

### LINKED ISSUE #432: Better Display impl for `grpcio::Error`

Ten0:
Helps #432 

This is an complementary approach to #433 for improving the `Display` impl for `grpcio::Error`, which requires less code. It has the additional effect of improving the `RpcStatusCode` and `RpcStatus` debug displays.

Note that this doesn't make #433 unnecessary: #433 still improves the `Display` implementation for `grpcio::Error` compared to this.

Once this is merged though, #433 can be expressed in an easier way.


BusyJay:
Now #433 is merged, I think you can delegate debug to display directly.

Ten0:
I also very often use `Debug`, be it for some structures that wrap a grpc error and derive `Debug` or because I `dbg!` some `RpcStatus` at some point, and I'm pretty sure it's more useful to see what the status code corresponds to than the internal structure of `RpcStatusCode`, which anyway has only one field, and it's private.

Do you think it's more useful to have the current `Debug` display, and if so, why?

BusyJay:
I think you get me wrong. I'm OK with the format you try to implement, I just suggest to implement it by delegating debug to display as they are essential the same code.

Ten0:
Oh, right, I need to update this PR based on the fact #443 is merged 🙂

Ten0:
Done

PR #440, opened 2020-03-10
Fix test_early_exit

### LINKED ISSUE #439: Cancel test met failure sometimes
**Describe the bug**
In `test_early_exit` test, a child thread could be detached implicitly. This child thread will use `unbounded_send`, which will unwrap an Error if the parent thread has ended.

**To Reproduce**
Make the integration test run long enough, as long as the integration test is not end 5s after the `test_early_exit test` is performed, this bug will appear.

hunterlxt:
Add judgement for the result of `unbounded_send()`

Close #439 


hunterlxt:
@overvenus PTAL

PR #445, opened 2020-03-19
*: clarify load balancing status

BusyJay:
Load balancing is supported in the grpc c core, but load report is not
implemented yet. It's hard to implement load report without census API.

Related grpc/grpc#22418.


hunterlxt:
Nice Job

PR #446, opened 2020-03-19
*: support unix domain socket

### LINKED ISSUE #296: add UNIX domain socket support
Support for using unix domain socket would be nice.

BusyJay:
It's supported by c core, we just need to allow it as parameter.

Close #296.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -64,7 +64,9 @@ where
 /// Given a host and port, creates a string of the form "host:port" or
 /// "[host]:port", depending on whether the host is an IPv6 literal.
 fn join_host_port(host: &str, port: u16) -> String {
-    if let Ok(ip) = host.parse::<IpAddr>() {
+    if host.starts_with("unix:") {
+        host.to_owned()
+    } else if let Ok(ip) = host.parse::<IpAddr>() {
         format!("{}\0", SocketAddr::new(ip, port))
     } else {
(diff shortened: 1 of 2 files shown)


tests-and-examples/tests/cases/misc.rs
```diff
@@ -129,3 +129,38 @@ fn test_soundness() {
     }
     assert_eq!(counter.load(Ordering::SeqCst), 9000);
 }
+
+#[test]
+fn test_uds() {
```
nrc:
```suggestion
fn test_unix_domain_service() {
```

BusyJay:
There should be bugs with rustc nightly. I can compile the PR with last nightly version.

BusyJay:
@nrc PTAL

src/server.rs
```diff
@@ -64,7 +64,9 @@ where
 /// Given a host and port, creates a string of the form "host:port" or
 /// "[host]:port", depending on whether the host is an IPv6 literal.
 fn join_host_port(host: &str, port: u16) -> String {
-    if let Ok(ip) = host.parse::<IpAddr>() {
+    if host.starts_with("unix:") {
```
hicqu: The return value must be nul-terminated, so how about use `CString` instead?

BusyJay: There may be two allocations, one is for format, second is for cstring.

PR #452, opened 2020-03-31
grpc-sys: downgrade bindgen

BusyJay:
Due to rust-lang/cargo#5237, upgrading bindgen is not backward
compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-# 0.5.1 - 2020-03-20
+# grpcio-sys 0.5.2 - 2020-03-31
+
+- Downgrade bindgen version to be backward compatible. (#452)
+
+# 0.5.1 - 2020-03-30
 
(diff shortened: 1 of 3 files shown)


hunterlxt:
Can you link to some known related issues we found here?

BusyJay:
> Can you link to some known related issues we found here?

Yes, it was provided in the PR description.

heftig:
I'm trying to upgrade a project from grpcio 0.4.7 and another dependency is already using bindgen 0.53.2. It seems my options right now are "don't upgrade" and "pin grpcio to 0.5.1".

heftig:
> bindgen should not be upgraded util a minor or major release

You didn't use either the 0.6 or the 0.7 release to bump the bindgen dependency.

BusyJay:
Indeed, I seem to forget about this. Can you send a PR so that I can get it merged when about releasing 0.8?

PR #456, opened 2020-04-04
Enable ALPN as default

hunterlxt:
NPN, the Next Protocol Negotiation extension, is specified by [a draft expired 6 years ago](https://tools.ietf.org/html/draft-agl-tls-nextprotoneg-04) and has been replaced by ALPN. So I suggest that we should support its successor ALPN to ensure successful connection to the servers with ALPN only.

Signed-off-by: TXXT <hunterlxt@live.com>

### DIFF
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -130,17 +130,6 @@ fn build_grpc(cc: &mut Build, library: &str) {
         if cfg!(feature = "openssl") {
             config.define("gRPC_SSL_PROVIDER", "package");
             config.define("EMBED_OPENSSL", "false");
-            // check if openssl provided from system support ALPN
-            if !cfg!(feature = "openssl-vendored")
-                && Build::new()
-                    .file("grpc/test/build/openssl-alpn.c")
-                    .cargo_metadata(false)
-                    .cpp(true)
-                    .try_compile("check_alpn")
-                    .is_err()
-            {
-                config.cxxflag("-DTSI_OPENSSL_ALPN_SUPPORT=0");
-            }
             setup_openssl(&mut config)
         } else if cfg!(feature = "secure") {
             third_party.extend_from_slice(&["boringssl/ssl", "boringssl/crypto"]);
@@ -261,7 +250,18 @@ fn setup_openssl(config: &mut Config) {
 }
 
 #[cfg(not(feature = "openssl-vendored"))]
-fn setup_openssl(_config: &mut Config) {}
+fn setup_openssl(config: &mut Config) {
+    // check if openssl provided from system support ALPN
+    if Build::new()
+        .file("grpc/test/build/openssl-alpn.c")
+        .cargo_metadata(false)
+        .cpp(true)
+        .try_compile("check_alpn")
+        .is_err()
+    {
+        config.cxxflag("-DTSI_OPENSSL_ALPN_SUPPORT=0");
+    }
+}
 
 fn get_env(name: &str) -> Option<String> {
     println!("cargo:rerun-if-env-changed={}", name);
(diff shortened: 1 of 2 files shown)


yiwu-arbug:
Fixing https://github.com/tikv/tikv/issues/7312

siddontang:
do it have compatibility problem?

hunterlxt:
> do it have compatibility problem?

If use the openssl vendored, it must have ALPN support. The openssl that comes with ubuntu14.04 or before does not have this.

BusyJay:
Is there any way to keep backward compatible? For example, toggle the flag according to openssl version?

yiwu-arbug:
> Is there any way to keep backward compatible? For example, toggle the flag according to openssl version?

ubuntu trusty is 6 years old. openssl 1.0.2 is 5 years old. Maybe we don't need to support such old systems?

hunterlxt:
IMO, the server side that does not support ALPN is almost non-existent. Considering ecosystem in grpc over tls, ALPN should be a mandatory option, not optional.

Similar discussions happened a few years ago, and multiple libraries remove support for NPN. Now only supporting NPN is not an good choice.

BusyJay:
It's about backward compatibility. If it's not backward compatible, the patch won't be released on 0.5.x.

grpc-sys/build.rs
```diff
@@ -130,6 +130,17 @@ fn build_grpc(cc: &mut Build, library: &str) {
...
+            // check if openssl provided from system support ALPN
+            if !cfg!(feature = "openssl-vendored")
+                && Build::new()
+                    .file("grpc/test/build/openssl-alpn.c")
+                    .cargo_metadata(false)
+                    .cpp(true)
+                    .try_compile("check_alpn")
+                    .is_err()
+            {
+                config.cxxflag("-DTSI_OPENSSL_ALPN_SUPPORT=0");
```
yiwu-arbug, line 142: should we fail to compile if openssl doesn't support ALPN?

grpc-sys/build.rs
```diff
@@ -130,10 +130,17 @@ fn build_grpc(cc: &mut Build, library: &str) {
         if cfg!(feature = "openssl") {
             config.define("gRPC_SSL_PROVIDER", "package");
             config.define("EMBED_OPENSSL", "false");
-            // Problem is: Ubuntu Trusty shipped with openssl 1.0.1f. Which doesn't
-            // support alpn. And Google's gRPC checks for support of ALPN in plane
-            // old Makefile, but not in CMake.
-            config.cxxflag("-DTSI_OPENSSL_ALPN_SUPPORT=0");
+            // check if openssl provided from system support ALPN
+            if !cfg!(feature = "openssl-vendored")
+                && Build::new()
```
BusyJay, line 135: It should be moved to `setup_openssl`.

grpc-sys/build.rs
```diff
@@ -254,7 +250,18 @@ fn setup_openssl(config: &mut Config) {
...
 
 #[cfg(not(feature = "openssl-vendored"))]
-fn setup_openssl(_config: &mut Config) {}
+fn setup_openssl(config: &mut Config) {
+    // check if openssl provided from system support ALPN
+    if Build::new()
+        .file("grpc/test/build/openssl-alpn.c")
+        .cargo_metadata(false)
+        .cpp(true)
+        .try_compile("check_alpn")
```
BusyJay, line 259: How about convert to `Tool` and then compile? Seems call `try_compile` directly will link the output to the crate.

hunterlxt: `cargo_metadata(false)` will prevent that thing

hunterlxt:
Seems that something wrong in xcode CI environment @BusyJay

grpc-sys/build.rs
BusyJay: Got it.

yiwu-arbug:
looks like travis failure is because there's too many warning logs. not sure if bumping travis xcode version will help.

Screen Shot 2020-04-10 at 1 31 14 PM

Issue #443, opened 2020-03-17
Support the "Richer error model"


kennytm:
**Is your feature request related to a problem? Please describe.**

According to https://grpc.io/docs/guides/error/#richer-error-model, gRPC supports [embedding `Any` message into the error reply](https://cloud.google.com/apis/design/errors#error_model). 

```proto
message Status {
  int32 code = 1;
  string message = 2;

  // Additional error information that the client code can use to handle
  // the error, such as retry delay or a help link.
  repeated google.protobuf.Any details = 3;
}
```

However, `grpcio` only supports an error message string.

https://github.com/tikv/grpc-rs/blob/d4ba053805dc984d96f075d788d03503e504a1c8/src/call/mod.rs#L154-L162

So if we really want to return structured error information, we need to use some bad code patterns like returning *success* and [include the error info in the *success* response](https://github.com/pingcap/kvproto/blob/2cf9a243b8d589f345de1dbaa9eeffec6afbdc06/proto/kvrpcpb.proto#L128-L133). 

```proto
message GetResponse {
    errorpb.Error region_error = 1;   // <-- wtf?
    KeyError error = 2;               // <-- ಠ_ಠ
    bytes value = 3;
    bool not_found = 4;
}
```

Note that 

> This richer error model is already supported in the C++, Go, Java, Python, and Ruby libraries, and at least the grpc-web and Node.js libraries have open issues requesting it. 

and also [`tower-grpc` exposes the array as opaque bytes](https://docs.rs/tower-grpc/0.1.1/tower_grpc/struct.Status.html#method.details).

**Describe the solution you'd like**

Support the "Richer error model". Add a new `additional_details: Vec<u8>` field with documentation how to encode/decode into protobuf-enabled types.

**Describe alternatives you've considered**

**Additional context**

BusyJay:
Related #284.

The proposal looks good. Do you want to bring it into grpc-rs?

BusyJay:
This will become a breaking change. If you want to land it on 0.5.x, we better use a different name than `details`. We can change the name to follow the convention in 0.6.x.

kennytm:
@BusyJay 

> Do you want to bring it into grpc-rs?

As long as TiKV is still using this package, yes.

> We can change the name to follow the convention in 0.6.x.

[...]

PR #457, opened 2020-04-04
Fetch google.rpc.Status proto if present

stuhood:
This allows us to get metadata from the server around _why_ a failure
happened.

We've been using this patch in an older version of the `grpcio` crate (https://github.com/pantsbuild/grpc-rs/commit/4dfafe9355dc996d7d0702e7386a6fedcd9734c0) for a while now to enable handling "both rpc status errors as well as message-inline errors" (see https://github.com/pantsbuild/pants/pull/6589).

### DIFF
--- a/src/call/mod.rs
+++ b/src/call/mod.rs
@@ -5,7 +5,7 @@ pub mod server;
 
 use std::fmt::{self, Debug, Display};
 use std::sync::Arc;
-use std::{ptr, slice};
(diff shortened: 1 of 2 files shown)


BusyJay:
This seems like a solution to #443 and #284. But it seems lack send side implement. And I'm not sure if it's the correct implement to still parse status_details to string, gRPC seems to treat them as bytes slice.

PR #461, opened 2020-05-07
*: bump 0.5.3

BusyJay:
This PR reverts future implement from master and bump v0.5.3.


hunterlxt:
why revert future implement ?

BusyJay:
Future will change the interfaces, should be landed at 0.6.x.

PR #468, opened 2020-06-12
Replace Spinlock with parking_lot::Mutex

hunterlxt:
After testing, the performance of parking_lot is better than `Spinlock` or `std::sync::Mutex`

Note: Only replace the lock in grpcio, the lock in the example remains the original usage


Cargo.toml
```diff
@@ -24,6 +24,7 @@ protobuf = { version = "2.0", optional = true }
 prost = { version = "0.6", optional = true }
 bytes = { version = "0.5", optional = true }
 log = "0.4"
+parking_lot = "0"
```
BusyJay, line 27: Should give it a version.

PR #469, opened 2020-06-17
Enhance sinks to make them batchable

### LINKED ISSUE #415: Enhance sinks to make them batchable
In the current implementation of `Sink`s in grpc-rs, every `start_send` will call `grpc_call_start_batch` to send one message. It makes grpc use much CPU. However, we can just put messages into buffer during calling `Sink::start_send`, but send them out when calling `Sink::poll_complete`. For both `Sink::send_all` or `Stream::forward`, `poll_complete` will be called after some `start_send`s, so it's ok. It should increase throughput, especially for overload cases.

hunterlxt:
close #415 

Optimize `send_all` using sink in grpc-rs. By buffering the message to be sent to determine whether there is still message in the stream, and add the `buffer_hint` flag to indicate follow-up messages.

After a simple test, this does improve the QPS by 50% when you use `send_all` method.

```
Before
======== test_client_stream_send_split
QPS: 13626.411713542033
======== test_client_stream_send_all
QPS: 13458.8795676736
======== test_server_stream_send_all
QPS: 21513.004479575207

After
======== test_client_stream_send_split
QPS: 13632.78241828707
======== test_client_stream_send_all
QPS: 20634.743684214573
======== test_server_stream_send_all
QPS: 21282.527286920154

Tcpdump capture file size (only send_all test)
Before：22M
After：2.6M
```

Related test code: https://github.com/hunterlxt/grpc-rs/blob/xt/bench-sink-over-master/tests-and-examples/examples/hello_world/client.rs

Additional benchmark with standard grpc driver (before vs after)
```
rust_protobuf_async_streaming_from_client_100msg_unconstrained
QPS: 12058.0
QPS: 41914.1
rust_protobuf_async_streaming_ping_pong
QPS: 18315.0
QPS: 20409.9
rust_protobuf_async_streaming_qps_unconstrained
QPS: 589875.8
QPS: 584359.6
```

### DIFF
--- a/benchmark/src/bench.rs
+++ b/benchmark/src/bench.rs
@@ -61,15 +61,8 @@ impl BenchmarkService for Benchmark {
         sink: ClientStreamingSink<SimpleResponse>,
     ) {
         let f = async move {
-            let mut req = SimpleRequest::default();
-            while let Some(r) = stream.try_next().await? {
-                req = r;
-            }
-            if req.get_response_size() > 0 {
-                sink.success(gen_resp(&req)).await?;
-            } else {
-                sink.success(Default::default()).await?;
-            }
+            let _sink = sink;
+            while stream.try_next().await?.is_some() {}
             Ok(())
         };
         let keep_running = self.keep_running.clone();
--- a/benchmark/src/bench.rs
+++ b/benchmark/src/bench.rs
@@ -61,8 +61,11 @@ impl BenchmarkService for Benchmark {
         sink: ClientStreamingSink<SimpleResponse>,
     ) {
         let f = async move {
-            let _sink = sink;
-            while stream.try_next().await?.is_some() {}
+            let mut req = SimpleRequest::default();
+            while let Some(r) = stream.try_next().await? {
+                req = r;
+            }
+            sink.success(gen_resp(&req));
             Ok(())
         };
         let keep_running = self.keep_running.clone();
--- a/benchmark/src/bench.rs
+++ b/benchmark/src/bench.rs
@@ -61,13 +61,13 @@ impl BenchmarkService for Benchmark {
         sink: ClientStreamingSink<SimpleResponse>,
     ) {
         let f = async move {
-            let mut resp: Option<SimpleResponse> = None;
-            while let Some(req) = stream.try_next().await? {
-                if resp.is_none() {
-                    resp = Some(gen_resp(&req));
-                }
+            let mut req = SimpleRequest::default();
+            while let Some(r) = stream.try_next().await? {
+                req = r;
             }
-            sink.success(resp.unwrap()).await?;
+            if req.get_response_size() > 0 {
+                sink.success(gen_resp(&req)).await?;
+            }
             Ok(())
         };
         let keep_running = self.keep_running.clone();
--- a/benchmark/src/bench.rs
+++ b/benchmark/src/bench.rs
@@ -7,6 +7,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use futures::prelude::*;
+use futures::sink::SinkExt;
 use grpc::{
     self, ClientStreamingSink, DuplexSink, MessageReader, Method, MethodType, RequestStream,
     RpcContext, RpcStatus, RpcStatusCode, ServerStreamingSink, ServiceBuilder, UnarySink,
@@ -56,12 +57,21 @@ impl BenchmarkService for Benchmark {
     fn streaming_from_client(
         &mut self,
         ctx: RpcContext,
-        _: RequestStream<SimpleRequest>,
+        mut stream: RequestStream<SimpleRequest>,
         sink: ClientStreamingSink<SimpleResponse>,
     ) {
-        let f = sink.fail(RpcStatus::new(RpcStatusCode::UNIMPLEMENTED, None));
+        let f = async move {
+            let mut resp: Option<SimpleResponse> = None;
+            while let Some(req) = stream.try_next().await? {
+                if resp.is_none() {
+                    resp = Some(gen_resp(&req));
+                }
+            }
+            sink.success(resp.unwrap()).await?;
+            Ok(())
+        };
         let keep_running = self.keep_running.clone();
-        spawn!(ctx, keep_running, "reporting unimplemented method", f)
+        spawn!(ctx, keep_running, "streaming from client", f)
     }
 
     fn streaming_from_server(
--- a/benchmark/src/bench.rs
+++ b/benchmark/src/bench.rs
@@ -67,6 +67,8 @@ impl BenchmarkService for Benchmark {
             }
             if req.get_response_size() > 0 {
                 sink.success(gen_resp(&req)).await?;
+            } else {
+                sink.success(Default::default()).await?;
             }
             Ok(())
         };
--- a/benchmark/src/client.rs
+++ b/benchmark/src/client.rs
@@ -182,6 +182,7 @@ struct RequestExecutor<B> {
     ctx: ExecutorContext<B>,
     client: Arc<BenchmarkServiceClient>,
     req: SimpleRequest,
+    cfg: ClientConfig,
 }
 
 impl<B: Backoff + Send + 'static> RequestExecutor<B> {
@@ -190,6 +191,7 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
             ctx,
             client: Arc::new(BenchmarkServiceClient::new(channel)),
             req: gen_req(cfg),
+            cfg: cfg.clone(),
         }
     }
 
@@ -254,6 +256,38 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
         };
         spawn!(client, keep_running, "streaming ping pong", f);
     }
+
+    fn execute_stream_from_client(mut self) {
+        let client = self.client.clone();
+        let keep_running = self.ctx.keep_running.clone();
+
+        let mut send_data = vec![];
+        for _ in 0..self.cfg.get_messages_per_stream() {
+            send_data.push(self.req.clone());
+        }
+        let f = async move {
+            loop {
+                let (mut sender, receiver) = self.client.streaming_from_client().unwrap();
+                let latency_timer = Instant::now();
+                let send_stream = futures::stream::iter(send_data.clone());
+                sender
+                    .send_all(&mut send_stream.map(move |item| Ok((item, WriteFlags::default()))))
+                    .await?;
+                sender.close().await?;
+                receiver.await?;
+                self.ctx.observe_latency(latency_timer.elapsed());
+                let mut time = self.ctx.backoff_async();
+                if let Some(t) = &mut time {
+                    t.await;
+                }
+                if !self.ctx.keep_running() {
+                    break;
+                }
+            }
+            Ok(())
+        };
+        spawn!(client, keep_running, "streaming from client", f);
+    }
 }
 
 fn execute<B: Backoff + Send + 'static>(
@@ -272,7 +306,7 @@ fn execute<B: Backoff + Send + 'static>(
         ClientType::ASYNC_CLIENT => match cfg.get_rpc_type() {
             RpcType::UNARY => {
                 if cfg.get_payload_config().has_bytebuf_params() {
-                    panic!("only streaming is supported for generic service.");
+                    panic!("only ping pong streaming is supported for generic service.");
                 }
                 RequestExecutor::new(ctx, ch, cfg).execute_unary_async()
             }
@@ -282,6 +316,12 @@ fn execute<B: Backoff + Send + 'static>(
                 } else {
                     RequestExecutor::new(ctx, ch, cfg).execute_stream_ping_pong()
                 }
+            }
+            RpcType::STREAMING_FROM_CLIENT => {
+                if cfg.get_payload_config().has_bytebuf_params() {
+                    panic!("only ping pong streaming is supported for generic service.");
+                }
+                RequestExecutor::new(ctx, ch, cfg).execute_stream_from_client()
             }
             _ => unimplemented!(),
         },
--- a/benchmark/src/client.rs
+++ b/benchmark/src/client.rs
@@ -261,30 +261,33 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
         let keep_running = self.ctx.keep_running.clone();
 
         let f = async move {
-            let (mut sender, _receiver) = self.client.streaming_from_client().unwrap();
+            let (mut sender, receiver) = self.client.streaming_from_client().unwrap();
 
             let send_stream = Box::pin(stream::unfold(
-                (self, Instant::now()),
-                |(mut c, last_time)| async move {
-                    c.ctx.observe_latency(last_time.elapsed());
-                    let mut time = c.ctx.backoff_async();
-                    if let Some(t) = &mut time {
-                        t.await;
+                (self, true, Instant::now()),
+                |(mut c, init, last_time)| async move {
+                    if !init {
+                        c.ctx.observe_latency(last_time.elapsed());
+                        let mut time = c.ctx.backoff_async();
+                        if let Some(t) = &mut time {
+                            t.await;
+                        }
                     }
-
                     if c.ctx.keep_running() {
-                        Some((c.req.clone(), (c, Instant::now())))
+                        Some((c.req.clone(), (c, false, Instant::now())))
                     } else {
                         None
                     }
                 },
             ));
 
+            sender.enhance_batch(true);
             sender
                 .send_all(&mut send_stream.map(move |item| Ok((item, WriteFlags::default()))))
                 .await?;
 
             sender.close().await?;
+            receiver.await?;
             Ok(())
         };
         spawn!(client, keep_running, "streaming from client", f);
--- a/benchmark/src/client.rs
+++ b/benchmark/src/client.rs
@@ -8,6 +8,7 @@ use std::time::{Duration, Instant};
 
 use futures::channel::oneshot::{self, Receiver, Sender};
 use futures::prelude::*;
+use futures::stream;
 use grpcio::{
     CallOption, Channel, ChannelBuilder, Client as GrpcClient, EnvBuilder, Environment, WriteFlags,
 };
@@ -182,7 +183,6 @@ struct RequestExecutor<B> {
     ctx: ExecutorContext<B>,
     client: Arc<BenchmarkServiceClient>,
     req: SimpleRequest,
-    cfg: ClientConfig,
 }
 
 impl<B: Backoff + Send + 'static> RequestExecutor<B> {
@@ -191,7 +191,6 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
             ctx,
             client: Arc::new(BenchmarkServiceClient::new(channel)),
             req: gen_req(cfg),
-            cfg: cfg.clone(),
         }
     }
 
@@ -257,33 +256,35 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
         spawn!(client, keep_running, "streaming ping pong", f);
     }
 
-    fn execute_stream_from_client(mut self) {
+    fn execute_stream_from_client(self) {
         let client = self.client.clone();
         let keep_running = self.ctx.keep_running.clone();
 
-        let mut send_data = vec![];
-        for _ in 0..self.cfg.get_messages_per_stream() {
-            send_data.push(self.req.clone());
-        }
         let f = async move {
-            loop {
-                let (mut sender, receiver) = self.client.streaming_from_client().unwrap();
-                let latency_timer = Instant::now();
-                let send_stream = futures::stream::iter(send_data.clone());
-                sender
-                    .send_all(&mut send_stream.map(move |item| Ok((item, WriteFlags::default()))))
-                    .await?;
-                sender.close().await?;
-                receiver.await?;
-                self.ctx.observe_latency(latency_timer.elapsed());
-                let mut time = self.ctx.backoff_async();
-                if let Some(t) = &mut time {
-                    t.await;
-                }
-                if !self.ctx.keep_running() {
-                    break;
-                }
-            }
+            let (mut sender, _) = self.client.streaming_from_client().unwrap();
+
+            let send_stream = Box::pin(stream::unfold(
+                (self, Instant::now()),
+                |(mut c, last_time)| async move {
+                    c.ctx.observe_latency(last_time.elapsed());
+                    let mut time = c.ctx.backoff_async();
+                    if let Some(t) = &mut time {
+                        t.await;
+                    }
+
+                    if c.ctx.keep_running() {
+                        Some((c.req.clone(), (c, Instant::now())))
+                    } else {
+                        None
+                    }
+                },
+            ));
+
+            sender
+                .send_all(&mut send_stream.map(move |item| Ok((item, WriteFlags::default()))))
+                .await?;
+
+            sender.close().await?;
             Ok(())
         };
         spawn!(client, keep_running, "streaming from client", f);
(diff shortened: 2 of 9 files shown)


src/call/mod.rs
```diff
@@ -646,29 +648,20 @@ impl SinkBase {
         &mut self,
         call: &mut C,
         t: &T,
-        mut flags: WriteFlags,
+        flags: WriteFlags,
         ser: SerializeFn<T>,
     ) -> Result<()> {
-        // `start_send` is supposed to be called after `poll_ready` returns ready.
-        assert!(self.batch_f.is_none());
+        // If there is already a buffered message waiting to be sent, set `buffer_hint` to true to indicate
```
hicqu, line 654: Need to check `self.send_metadata` first.

hunterlxt: will be checked in the actual function.

https://github.com/tikv/grpc-rs/pull/469/files#diff-8c8308f2c854f39735f331e70f94d8d2R705-R708

BusyJay:
Cool, how does grpcio benchmark say?

src/call/mod.rs
hicqu: Yes, but I prefer to keep it into `start_send`, because I think it's more clear.

src/call/mod.rs
```diff
@@ -683,4 +676,46 @@ impl SinkBase {
...
+        }
+
+        let write_f = call.call(|c| {
+            c.call
+                .start_send_message(&self.buf, flags.flags, self.send_metadata)
+        })?;
+
+        self.send_metadata = false;
+        self.batch_f = Some(write_f);
+        self.buf.clear();
```
hicqu, line 717: Don't drop the `shrink` logic, because the buffer can grow too large.

hunterlxt:
> Cool, how does grpcio benchmark say?

The updated results are in the description. BTW, I read the benchmark code, the client does not use the `send_all` method.

BusyJay:
If adapted to send_all, how does it performs?

Since now std::future is used, async/await is the preferred style. It will be less useful if the optimization can't work with that.

hunterlxt:
> If adapted to send_all, how does it performs?
> 
> Since now std::future is used, async/await is the preferred style. It will be less useful if the optimization can't work with that.

`send_all` is a method of `SinkExt`. `futures::sink` and `std::future` do not conflict with each other. For users, the method of std::future is not directly used, but the method in `SinkExt` of `futures::sink`.

Above is the background. The semantics of `send()` is to send and ensure that this message is indeed successful. But if the user really wants to send a stream, he should choose the `send_all` method, and our sink has not been optimized for this.

BusyJay:
Can you give an example about how to utilize the optimization with async/await?

hunterlxt:
The previous benchmark in grpc-rs did not implement the `stream_from_client` bench. I added it and performed the bench. The result has been updated. cc @BusyJay

benchmark/src/bench.rs
```diff
@@ -56,12 +57,21 @@ impl BenchmarkService for Benchmark {
     fn streaming_from_client(
         &mut self,
         ctx: RpcContext,
-        _: RequestStream<SimpleRequest>,
+        mut stream: RequestStream<SimpleRequest>,
         sink: ClientStreamingSink<SimpleResponse>,
     ) {
-        let f = sink.fail(RpcStatus::new(RpcStatusCode::UNIMPLEMENTED, None));
+        let f = async move {
```
hicqu, line 63: Seems it's a duplex streaming? So it's better to send a response for every request?

src/call/mod.rs
```diff
@@ -683,4 +686,44 @@ impl SinkBase {
...
+        &mut self,
+        cx: &mut Context,
+        call: &mut C,
+    ) -> Poll<Result<()>> {
+        if self.batch_f.is_some() {
+            ready!(self.poll_ready(cx)?);
+        }
+        if self.buf_flags.is_some() {
+            // Set `buffer_hint` to false to indicate that this is the last message.
+            self.start_send_buffer_message(false, call)?;
```
hicqu, line 701: What if all messages are pushed into the sink with `buffer hint = false`?

benchmark/src/bench.rs
hunterlxt: // Single-sided unbounded streaming from client to server
// The server returns the client payload as-is once the client does WritesDone
rpc StreamingFromClient(stream SimpleRequest) returns (SimpleResponse);

src/call/mod.rs
hunterlxt: Fixed

src/call/mod.rs
```diff
@@ -646,29 +654,28 @@ impl SinkBase {
...
-        // NOTE: Content of `self.buf` is copied into grpc internal.
-        if self.buf.capacity() > BUF_SHRINK_SIZE {
-            self.buf.truncate(BUF_SHRINK_SIZE);
-            self.buf.shrink_to_fit();
+        // If there is already a buffered message waiting to be sent, set `buffer_hint` to true to indicate
+        // that this is not the last message.
+        if self.buf_flags.is_some() {
+            // `start_send` is supposed to be called after `poll_ready` returns ready.
+            assert!(self.batch_f.is_none());
+            self.start_send_buffer_message(true, call)?;
```
BusyJay, line 674: Any tests?

src/call/mod.rs
```diff
@@ -683,4 +690,43 @@ impl SinkBase {
...
+        buffer_hint: bool,
+        call: &mut C,
+    ) -> Result<()> {
+        let mut flags = self.buf_flags.clone().unwrap();
+        flags = flags.buffer_hint(buffer_hint);
+        let write_f = call.call(|c| {
+            c.call
+                .start_send_message(&self.buf, flags.flags, self.send_metadata)
+        })?;
+        self.batch_f = Some(write_f);
```
BusyJay, line 722: This can go wrong. Let's say user sends [(msg1, not_bufferred), (msg2, bufferred)], old implement will only keep msg2 in buffer, but new implement will keep both.

hunterlxt: This is the meaning of this PR, buffer the messages in a stream as much as possible

src/call/mod.rs
hunterlxt: The integration test is already included in the example of the route guide, if I just migrate the examples to the tests, I think it is not necessary

src/call/client.rs
```diff
@@ -331,6 +331,10 @@ impl<Req> StreamingCallSink<Req> {
         }
     }
 
+    pub fn enable_batch(&mut self, flag: bool) {
```
hicqu, line 334: How about `enhance_batch`? And please document the enhanced strategy detailly.

hicqu:
rest LGTM.

src/call/mod.rs
```diff
@@ -616,41 +616,50 @@ impl WriteFlags {
...
     }
 }
 
 /// A helper struct for constructing Sink object for batch requests.
+///
+/// The sink works in normal mode by default, that is, `start_send` always sends message immediately.
+/// But when the `enhance_buffer_strategy` is enabled, the message to be sent will be processed
+/// according to the following process:
+/// Set the `buffer_hint` of the non-end message in the stream to true, and set the `buffer_hint` of the
+/// last message to true in `poll_flush`, so that the previously bufferd messages will be sent out.
```
hicqu: s/to true/to false

hunterlxt: This is an outdated commit.

benchmark/src/bench.rs
```diff
@@ -56,12 +57,21 @@ impl BenchmarkService for Benchmark {
...
+            let mut resp: Option<SimpleResponse> = None;
+            while let Some(req) = stream.try_next().await? {
```
BusyJay, line 65: Can you reference the original implement?

src/call/client.rs
```diff
@@ -331,6 +331,15 @@ impl<Req> StreamingCallSink<Req> {
         }
     }
 
+    /// By default the sink works in normal mode, that is, `start_send` always starts to send message
+    /// immediately. But when the `enhance_batch` is enabled, the stream will be batched together as
+    /// much as possible. The specific rule is listed below:
+    /// Set the `buffer_hint` of the non-end message in the stream to true, and set the `buffer_hint` of
+    /// the last message to false in `poll_flush`, so that the previously bufferd messages will be sent out.
```
BusyJay, line 338: The implement doesn't match the description. The implement sets buffer hint to false only when there is at least a false flag.

benchmark/src/bench.rs
hunterlxt: grpc/test/cpp/qps/server_sync.cc: 120

```c++
  static Status ClientPull(ServerContext* /*context*/, R* stream,
                           SimpleResponse* response) {
    SimpleRequest request;
    while (stream->Read(&request)) {
    }
    if (request.response_size() > 0) {
      if (!Server::SetPayload(request.response_type(), request.response_size(),
                              response->mutable_payload())) {
        return Status(grpc::StatusCode::INTERNAL, "Error creating payload.");
      }
    }
    return Status::OK;
  }
```

src/call/client.rs
hunterlxt: Fixed

benchmark/src/bench.rs
BusyJay: Then the implement is wrong here, cpp constructs the response using the last request.

hunterlxt: I'll fix it

benchmark/src/bench.rs
```diff
@@ -56,12 +57,21 @@ impl BenchmarkService for Benchmark {
...
-        let f = sink.fail(RpcStatus::new(RpcStatusCode::UNIMPLEMENTED, None));
+        let f = async move {
+            let mut req = SimpleRequest::default();
+            while let Some(r) = stream.try_next().await? {
+                req = r;
+            }
+            if req.get_response_size() > 0 {
+                sink.success(gen_resp(&req)).await?;
+            }
+            Ok(())
```
BusyJay, line 71: When response size is zero, cpp will still respond a default response.

benchmark/src/client.rs
```diff
@@ -254,6 +256,38 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
         };
         spawn!(client, keep_running, "streaming ping pong", f);
     }
+
+    fn execute_stream_from_client(mut self) {
```
BusyJay, line 260: Can you reference the original implement?

src/call/client.rs
```diff
@@ -331,6 +331,16 @@ impl<Req> StreamingCallSink<Req> {
         }
     }
 
+    /// By default the sink works in normal mode, that is, `start_send` always starts to send message
+    /// immediately. But when the `enhance_batch` is enabled, the stream will be batched together as
+    /// much as possible. The specific rule is listed below:
```
BusyJay, line 336:
```suggestion
    /// much as possible. The specific rules are listed below:
```

src/call/mod.rs
```diff
@@ -616,59 +616,87 @@ impl WriteFlags {
...
     /// Get whether compression is disabled.
-    pub fn get_force_no_compress(self) -> bool {
+    pub fn get_force_no_compress(&self) -> bool {
         (self.flags & grpc_sys::GRPC_WRITE_NO_COMPRESS) != 0
     }
 }
 
 /// A helper struct for constructing Sink object for batch requests.
+///
+/// By default the sink works in normal mode, that is, `start_send` always starts to send message
```
BusyJay, line 631: Why paste the comment here?

src/call/mod.rs
```diff
@@ -616,59 +616,87 @@ impl WriteFlags {
...
 struct SinkBase {
+    // Batch job to be executed in `poll_ready`.
     batch_f: Option<BatchFuture>,
-    buf: Vec<u8>,
     send_metadata: bool,
+    // Flag to indicate if enhance batch strategy. This behavior will modify the `buffer_hint` to batch
+    // messages as much as possible.
+    enhance_buffer_strategy: bool,
+    // Buffer used to store the data to be sent, send out the last data in this round of `start_send`.
+    // Note: only used in enhanced buffer strategy.
```
BusyJay, line 645: I think it's used no matter what buffer strategy is used.

src/call/mod.rs
```diff
@@ -616,59 +616,87 @@ impl WriteFlags {
...
     fn start_send<T, C: ShareCallHolder>(
         &mut self,
         call: &mut C,
         t: &T,
-        mut flags: WriteFlags,
+        flags: WriteFlags,
         ser: SerializeFn<T>,
     ) -> Result<()> {
-        // `start_send` is supposed to be called after `poll_ready` returns ready.
-        assert!(self.batch_f.is_none());
```
BusyJay, line 653: Why removing the assert?

src/call/mod.rs
```diff
@@ -683,4 +711,46 @@ impl SinkBase {
...
+        let write_f = call.call(|c| {
+            c.call
+                .start_send_message(&self.buffer, flags.flags, self.send_metadata)
+        })?;
+        self.batch_f = Some(write_f);
+        self.buf_flags.take();
+        // NOTE: Content of `self.buf` is copied into grpc internal.
+        self.buffer.clear();
+        if self.buffer.capacity() > BUF_SHRINK_SIZE {
+            self.buffer.truncate(BUF_SHRINK_SIZE);
```
BusyJay, line 751: Because it's cleared at L749, so truncate will do nothing, and `shrink_to_fit` may shrink it to 0.

src/call/mod.rs
hunterlxt: I think the assert should be placed as close as possible to the function sending the message. So move it into `start_send_buffer_message`

benchmark/src/client.rs
hunterlxt:
```c++
  bool ThreadFuncImpl(HistogramEntry* entry, size_t thread_idx) override {
    // Figure out how to make histogram sensible if this is rate-paced
    if (!WaitToIssue(thread_idx)) {
      return true;
    }
    GPR_TIMER_SCOPE("SynchronousStreamingFromClientClient::ThreadFunc", 0);
    if (stream_[thread_idx]->Write(request_)) {
      double now = UsageTimer::Now();
      entry->set_value((now - last_issue_[thread_idx]) * 1e9);
      last_issue_[thread_idx] = now;
      return true;
    }
    stream_[thread_idx]->WritesDone();
    FinishStream(entry, thread_idx);
    auto* stub = channels_[thread_idx % channels_.size()].get_stub();
    std::lock_guard<std::mutex> l(stream_mu_[thread_idx]);
    if (!shutdown_[thread_idx].val) {
      stream_[thread_idx] = stub->StreamingFromClient(&context_[thread_idx],
                                                      &responses_[thread_idx]);
    } else {
      stream_[thread_idx].reset();
      return false;
    }
    return true;
  }
```
grpc/test/cpp/qps/client_sync.cc line:302~327

hunterlxt: In fact, when I wrote bench at the time, I did not fully refer to the implementation of grpc-cpp

src/call/mod.rs
hunterlxt: Yes, I removed `truncate` and if it grows up to `BUF_SHRINK_SIZE`, shrink can do some cleanup.

BusyJay: This is wrong. The whole point of buffer is to reduce allocation, this implement will make it allocate more.

benchmark/src/client.rs
BusyJay: > ...I did not fully refer to the implementation of grpc-cpp

You should. It's expected to work across implement.

benchmark/src/client.rs
```diff
@@ -254,6 +255,40 @@ impl<B: Backoff + Send + 'static> RequestExecutor<B> {
...
+        let client = self.client.clone();
+        let keep_running = self.ctx.keep_running.clone();
+
+        let f = async move {
+            let (mut sender, _receiver) = self.client.streaming_from_client().unwrap();
+
+            let send_stream = Box::pin(stream::unfold(
+                (self, Instant::now()),
+                |(mut c, last_time)| async move {
+                    c.ctx.observe_latency(last_time.elapsed());
```
BusyJay, line 269: It should observe only after last is sent out.

benchmark/src/bench.rs
```diff
@@ -56,12 +57,16 @@ impl BenchmarkService for Benchmark {
     fn streaming_from_client(
         &mut self,
         ctx: RpcContext,
-        _: RequestStream<SimpleRequest>,
+        mut stream: RequestStream<SimpleRequest>,
         sink: ClientStreamingSink<SimpleResponse>,
     ) {
-        let f = sink.fail(RpcStatus::new(RpcStatusCode::UNIMPLEMENTED, None));
+        let f = async move {
+            let _sink = sink;
```
BusyJay, line 64: I think it needs to respond in the original implement.

src/call/mod.rs
```diff
@@ -616,59 +616,78 @@ impl WriteFlags {
     }
 
     /// Get whether buffer hint is enabled.
-    pub fn get_buffer_hint(self) -> bool {
+    pub fn get_buffer_hint(&self) -> bool {
```
BusyJay, line 619: Why change?

src/call/mod.rs
```diff
@@ -683,4 +702,46 @@ impl SinkBase {
...
+    fn poll_flush<C: ShareCallHolder>(
+        &mut self,
+        cx: &mut Context,
+        call: &mut C,
+    ) -> Poll<Result<()>> {
+        if self.batch_f.is_some() {
+            ready!(self.poll_ready(cx)?);
+        }
+        if self.buf_flags.is_some() {
+            self.start_send_buffer_message(self.last_buf_hint, call)?;
```
BusyJay, line 716: So the batch will not work again even after flush if it has been set to false in the history?

src/call/server.rs
```diff
@@ -424,6 +424,16 @@ macro_rules! impl_stream_sink {
                 }
             }
 
+            /// By default the sink works in normal mode, that is, `start_send` always starts to send message
+            /// immediately. But when the `enhance_batch` is enabled, the stream will be batched together as
+            /// much as possible. The specific rules are listed below:
+            /// Set the `buffer_hint` of the non-end message in the stream to true. And set the `buffer_hint`
+            /// of the last message to false in `poll_flush` only when there is at least one message with the
+            /// `buffer_hint` false, so that the previously bufferd messages will be sent out.
```
BusyJay, line 432:
```suggestion
            /// By default it always sends messages with their configured buffer hint. But when the
            /// `enhance_batch` is enabled, messages will be batched together as many as possible.
            /// The rules are listed as below:
            /// - All messages except the last one will be sent with `buffer_hint` set to true.
            /// - The last message will also be sent with `buffer_hint` set to true unless all messages are
            ///    offered with buffer hint set to true.
            ///
            /// No matter `enhance_batch` is true or false, it's recommended to follow the contract of
            /// Sink and call `poll_flush` to ensure messages are handled by gRPC C Core.
```

src/call/mod.rs
hunterlxt: Whether batch is enabled depends only on the current `enhance_buffer_strategy`

src/call/mod.rs
BusyJay: No you don't. WriteFlags is copy and fits in register. It doesn't need to be accessed by pointer indirectly.

src/call/mod.rs
BusyJay: No, once `last_buf_hint` becomes false, it will not become true again.

hunterlxt: you are right, but even if this hint is always false, it will not affect the batch, but it will affect the last message in the stream. (For example, if the user wants a stream of messages to be buffered, the last one will still not be bufferd)

src/call/server.rs
```diff
@@ -424,6 +424,19 @@ macro_rules! impl_stream_sink {
                 }
             }
 
+            /// By default it always sends messages with their configured buffer hint. But when the
+            /// `enhance_batch` is enabled, messages will be batched together as many as possible.
+            /// The rules are listed as below:
+            /// - All messages except the last one will be sent with `buffer_hint` set to true.
+            /// - The last message will also be sent with `buffer_hint` set to true unless all messages are
+            ///    offered with buffer hint set to true.
```
BusyJay, line 432:
```suggestion
            /// - The last message will also be sent with `buffer_hint` set to true unless any message is
            ///    offered with buffer hint set to false.
```

src/call/client.rs
```diff
@@ -331,6 +331,16 @@ impl<Req> StreamingCallSink<Req> {
         }
     }
 
+    /// By default the sink works in normal mode, that is, `start_send` always starts to send message
```
BusyJay, line 334: The comment needs to be updated too.

tests-and-examples/tests/cases/stream.rs
```diff
@@ -0,0 +1,126 @@
...
+            let mut p = Point::default();
+            p.set_longitude(i);
+            send_data.push(p);
+        }
+        let send_stream = futures::stream::iter(send_data);
+        sink.enhance_batch(true);
+        sink.send_all(&mut send_stream.map(move |item| Ok((item, WriteFlags::default()))))
+            .await
+            .unwrap();
+        sink.close().await.unwrap();
```
BusyJay, line 103: The test can't test if batch sets the last message to buffer correctly because `close` will always flush all messages. Better move it after L105 and make `record_route` returns immediately after handling given limited requests.

hunterlxt: adjusted test logic

tests-and-examples/tests/cases/stream.rs
```diff
@@ -0,0 +1,148 @@
...
+                .unwrap_err();
+            sink.close().await.unwrap();
+            rx.recv_timeout(std::time::Duration::from_secs(1)).unwrap();
+        };
+        let recv_msg_task = async move {
+            let summary = receiver.await.unwrap();
+            tx.send(()).unwrap();
+            assert_eq!(summary.get_point_count(), MESSAGE_NUM);
+        };
+        join!(close_sink_task, recv_msg_task);
```
BusyJay, line 145: The case is very hacky. It relies on scheduler polls future in FIFO order. A better approach is use futures channel instead of blocking.

hunterlxt: ou can adjust the sequence of close_sink_task and recv_msg_task and it doesn't matter. If use futures channel, this case cannot prove that all msgs will be buffered util sink.close() is called

BusyJay: You can combine channel with timer to get `recv_timeout`.

hunterlxt:
@BusyJay @hicqu PTAL

tests-and-examples/tests/cases/stream.rs
```diff
@@ -0,0 +1,161 @@
...
+        .build()
+        .unwrap();
+    server.start();
+    let port = server.bind_addrs().next().unwrap().1;
+    let ch = ChannelBuilder::new(env).connect(&format!("127.0.0.1:{}", port));
+    let client = RouteGuideClient::new(ch);
+
+    let exec_test_f = async move {
+        // Test for send all disable batch
+        {
```
BusyJay, line 89: We usually use blank line between cases instead of surrounding them with brackets.

tests-and-examples/tests/cases/stream.rs
```diff
@@ -0,0 +1,161 @@
...
+            let close_sink_task = async move {
+                rx.try_next().unwrap().unwrap();
+                Delay::new(std::time::Duration::from_secs(1)).await;
+                rx.try_next().unwrap_err();
+                sink.close().await.unwrap();
+                Delay::new(std::time::Duration::from_secs(1)).await;
+                rx.try_next().unwrap();
+            };
+            let recv_msg_task = async move {
+                tx.send(()).await.unwrap();
```
BusyJay, line 152: I think L152 and L144 can be removed.

PR #474, opened 2020-07-15
Add blocking callback to EnvBuilder

hunterlxt:
This allows users to manipulate the thread pool created by gRPC. They can add some custom code like print logs for newly created threads and reclaim some resources or record when exiting the thread.

Note: It is recommended to PR pick to 0.5.x


BusyJay:
It doesn't have to be joined to execute `before_stop` hook.

PR #475, opened 2020-07-15
Pick Add blocking callback to EnvBuilder (#474)

hunterlxt:

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -27,6 +27,6 @@ protobuf = "2"
 lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
-protobuf-build = { version = "0.11", default-features = false }
+protobuf-build = { version = "=0.11.3", default-features = false }
 grpcio-compiler = { path = "../compiler", version = "0.5.0", default-features = false }
 walkdir = "2.2"
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -28,5 +28,5 @@ lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
 protobuf-build = { version = "=0.11.3", default-features = false }
-grpcio-compiler = { path = "../compiler", version = "0.5.0", default-features = false }
+grpcio-compiler = { version = "=0.5.0", default-features = false }
 walkdir = "2.2"
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -28,5 +28,5 @@ lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
 protobuf-build = { version = "=0.11.3", default-features = false }
-grpcio-compiler = { version = "=0.5.0", default-features = false }
+grpcio-compiler = { path = "../compiler", version = "0.5.0", default-features = false }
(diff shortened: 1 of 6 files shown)


hunterlxt:
Windows build error is not related to this PR

BusyJay:
Better fix it before releasing a new version.

proto/Cargo.toml
```diff
@@ -27,6 +27,6 @@ protobuf = "2"
 lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
-protobuf-build = { version = "0.11", default-features = false }
+protobuf-build = { version = "=0.11.3", default-features = false }
```
BusyJay, line 30: Is it possible to limiting the grpcio-compiler version?

hunterlxt: Done

hunterlxt:
@hicqu @BusyJay windows build error has been fixed.

proto/Cargo.toml
```diff
@@ -27,6 +27,6 @@ protobuf = "2"
 lazy_static = { version = "1.3", optional = true }
 
 [build-dependencies]
-protobuf-build = { version = "0.11", default-features = false }
-grpcio-compiler = { path = "../compiler", version = "0.5.0", default-features = false }
+protobuf-build = { version = "=0.11.3", default-features = false }
```
BusyJay, line 30: Why still changing this?

proto/Cargo.toml
```diff
@@ -27,6 +27,6 @@ protobuf = "2"
...
+grpcio-compiler = { version = "=0.5.0", default-features = false }
```
BusyJay, line 31: This is wrong. It prevents from updating and testing local compiler.

hunterlxt: protobuf-build use grpcio-compiler from crates.io and locked it. So you mean patch grpcio-compiler with local crate in root cargo.toml?

proto/Cargo.toml
hunterlxt: protobuf-build must be 0.11.3, or it will produce code with std::future

BusyJay: Why? Isn't it constraint by L31?

proto/Cargo.toml
BusyJay: It can be a solution.

proto/Cargo.toml
hunterlxt: yes, cargo.lock showed that there are 2 grpcio-compiler and I'll patch it later.

hunterlxt:
@BusyJay It has been patched but 0.11.3 must still be restricted because 0.11.4 uses grpcio-compiler 0.6.0. And I suggest that protobuf-build use new version num like 0.12 because 0.11.3 and 0.11.4 are very different.

BusyJay:
Then patch configuration is useless.

hunterlxt:
@BusyJay Why? I did see that grpcio-compiler only appeared once in Cargo.lock

BusyJay:
I think whether there are two versions depends on whether protobuf-build is pinned instead of patch. Deleting patch should make no difference.

PR #481, opened 2020-08-24
*: use grpc_slice to avoid copy

BusyJay:
Use grpc_slice to reduce memory copy. If the messages are small,
this should reduce allocations.

### DIFF
--- a/src/buf.rs
+++ b/src/buf.rs
@@ -62,6 +62,11 @@ impl GrpcSlice {
         GrpcSlice::from_static_slice(s.as_bytes())
     }
 
+    /// Checks whether the slice stores bytes inline.
+    pub fn is_inline(&self) -> bool {
+        self.0.refcount.is_null()
+    }
+
     /// Reallocates current slice with given capacity.
     ///
     /// The length of returned slice is the exact same as given cap.
--- a/src/buf.rs
+++ b/src/buf.rs
@@ -7,6 +7,9 @@ use std::fmt::{self, Debug, Formatter};
 use std::io::{self, BufRead, Read};
 use std::mem::{self, ManuallyDrop, MaybeUninit};
 
+const INLINED_SIZE: usize = mem::size_of::<libc::size_t>() + mem::size_of::<*mut u8>() - 1
+    + mem::size_of::<*mut libc::c_void>();
+
 /// A convenient rust wrapper for the type `grpc_slice`.
 ///
 /// It's expected that the slice should be initialized.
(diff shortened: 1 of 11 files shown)


src/buf.rs
```diff
@@ -58,6 +61,41 @@ impl GrpcSlice {
...
+        if cap <= INLINED_SIZE {
+            // Only inlined slice can be reused safely.
+            if !self.0.refcount.is_null() {
+                *self = GrpcSlice::default();
+            }
+            self.0.data.inlined.length = cap as u8;
+            std::slice::from_raw_parts_mut(
+                self.0.data.inlined.bytes.as_mut_ptr() as *mut MaybeUninit<u8>,
+                cap,
+            )
```
hunterlxt, line 87: can you explain the function with more details. I don't understand why to compare `cap` with `INLINED_SIZE`. And what does it mean that refcount is null.

BusyJay: Check out: https://docs.rs/grpcio-sys/0.6.0/grpcio_sys/struct.grpc_slice.html.

src/buf.rs
```diff
@@ -58,6 +63,41 @@ impl GrpcSlice {
...
+    }
+
+    /// Reallocates current slice with given capacity.
+    ///
+    /// The length of returned slice is the exact same as given cap.
+    ///
+    /// ## Safety
+    ///
+    /// Caller is expected to initialize all available bytes to guarantee safety of this slice.
+    pub unsafe fn realloc(&mut self, cap: usize) -> &mut [MaybeUninit<u8>] {
```
hicqu, line 79: Seems `self` will be replaced. by a new one. So is `self.0.refcount.sub(1)` necessary?

BusyJay: It's done by `Drop` automatically.

src/buf.rs
```diff
@@ -58,6 +63,41 @@ impl GrpcSlice {
...
+            if !self.0.refcount.is_null() {
+                *self = GrpcSlice::default();
+            }
+            self.0.data.inlined.length = cap as u8;
+            std::slice::from_raw_parts_mut(
+                self.0.data.inlined.bytes.as_mut_ptr() as *mut MaybeUninit<u8>,
+                cap,
+            )
+        } else {
+            *self = GrpcSlice(grpcio_sys::grpc_slice_malloc_large(cap));
```
hicqu, line 91: Seems there is still no `capacity` in `grpc_slice`?

BusyJay: Indeed, `capacity == len`. That's why comment says "Caller is expected to initialize all available bytes to guarantee safety of this slice."

src/codec.rs
```diff
@@ -47,15 +55,22 @@ pub mod pb_codec {
...
     use prost::Message;
 
     use super::MessageReader;
+    use crate::buf::GrpcSlice;
     use crate::error::Result;
 
     #[inline]
-    pub fn ser<M: Message, B: BufMut>(msg: &M, buf: &mut B) {
-        msg.encode(buf).expect("Writing message to buffer failed");
+    pub fn ser<M: Message>(msg: &M, buf: &mut GrpcSlice) {
```
hicqu, line 65: How about return bytes? So that we can remamber this where `ser` is called. And then we can add `set_length` to `GrpcSlice` to reduce more allocs.

hicqu (approved):

hunterlxt (approved):

PR #486, opened 2020-09-21
*: bump 0.7.0

hunterlxt:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# 0.7.0 - 2020-11-02
+
+- Add blocking callback to `EnvBuilder` (#474)
+- Enhance sinks to make them batchable (#469)
+- Remove `rustfmt_skip` attribute since it is unstable (#479)
+- Use `grpc_slice` to reduce memory copy (#481)
+- Fix the bug that server cannot shutdown itself when drop (#484)
+- Add methods for channels from file descriptors (#488)
+- Update gRPC C core to 1.33.1 (#492)
+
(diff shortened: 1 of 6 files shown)


youjiali1995 (dismissed):

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.6.1 - 2020-09-21
+
+- Add blocking callback to `EnvBuilder` (#474)
```
BusyJay, line 3: Change in rust-compiler is missing.

compiler/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio-compiler"
-version = "0.6.0"
+version = "0.6.1"
```
BusyJay, line 3: Actually the changes is incompatible as it changes C interface. So grpcio 0.6.0 can't compile with grpcio-compiler 0.6.1. Maybe we should use a separate branch instead.

grpc-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio-sys"
-version = "0.6.0"
+version = "0.6.1"
```
BusyJay, line 3: Actually the changes is incompatible as it changes C interface. So grpcio 0.6.0 can't compile with grpcio-compiler 0.6.1. Maybe we should use a separate branch instead.

BusyJay (approved):

skyzh (approved):

PR #488, opened 2020-09-29
Add methods for channels from file descriptors

hcbdfinity:
Expose methods to allow creating server and client channels from existing
file descriptors.


src/channel.rs
```diff
@@ -442,6 +442,23 @@ impl ChannelBuilder {
         let channel =
             unsafe { grpc_sys::grpc_insecure_channel_create(addr_ptr, args.args, ptr::null_mut()) };
 
+        self.connect_from_raw_channel(channel)
+    }
+
+    /// Build an insecure [`Channel`] taking over an established connection from
+    /// a file descriptor.
+    pub fn connect_from_fd(mut self, target: &str, fd: ::std::os::raw::c_int) -> Channel {
```
BusyJay, line 450: It should be `unsafe` and unix only.

src/channel.rs
```diff
@@ -442,6 +442,23 @@ impl ChannelBuilder {
...
+        let target = CString::new(target).unwrap();
+        let target_ptr = target.as_ptr();
+        let channel =
+            unsafe { grpc_sys::grpc_insecure_channel_create_from_fd(target_ptr, fd, args.args) };
+
+        self.connect_from_raw_channel(channel)
+    }
+
+    /// Build a channel from a raw underlying core grpc channel
+    pub fn connect_from_raw_channel(self, channel: *mut grpc_channel) -> Channel {
```
BusyJay, line 461: Unnecessary function.

src/server.rs
```diff
@@ -561,6 +561,18 @@ impl Server {
     pub fn bind_addrs(&self) -> impl ExactSizeIterator<Item = (&String, u16)> {
         self.core.binders.iter().map(|b| (&b.host, b.port))
     }
+
+    /// Add an rpc channel for an established connection represented as a file
+    /// descriptor.
+    pub fn add_insecure_channel_from_fd(&self, fd: ::std::os::raw::c_int) {
```
BusyJay, line 567: Should be unsafe and unix only.

src/channel.rs
BusyJay: Should explain what's target and who is responsible to open and close fd.

hcbdfinity: thanks -- will amend doc, but how do I mark it "posix-only"?

src/channel.rs
hcbdfinity: no, this function is quite useful actually -- it allows binding custom rpc channels created by other means (e.g. third party custom transport implementations), maybe this warrants more documentation? also, might need marking as unsafe

hcbdfinity:
Thanks for review, amended documentation and "unsafe" annotations -- please advice on how I could mark the function in question as "posix-only", I actually do not know how to do that besides adding a comment.

src/channel.rs
BusyJay: Can you give more details about the use case? Is it a real use case?

src/channel.rs
BusyJay: `#[cfg(unix)]`

src/channel.rs
hcbdfinity: Might be a 'fringe' use case, although it is 'real' for me -- you basically always need that when you intend to use anything besides 'plain' sockets for transport.

src/channel.rs
hcbdfinity: thanks, added as attribute

src/channel.rs
```diff
@@ -442,6 +442,30 @@ impl ChannelBuilder {
...
+        let channel = grpc_sys::grpc_insecure_channel_create_from_fd(target_ptr, fd, args.args);
+
+        self.connect_from_raw_channel(channel)
+    }
+
+    /// Build a channel from a raw underlying core grpc channel. This allows to
+    /// establish an RPC connection using custom channel implementations.
+    /// Takes over ownership of the given channel object and will close it
+    /// when the connection is closed.
+    pub unsafe fn connect_from_raw_channel(self, channel: *mut grpc_channel) -> Channel {
```
BusyJay, line 468: You can mark `Channel::new` as public and unsafe instead.

src/server.rs
```diff
@@ -561,6 +561,13 @@ impl Server {
     pub fn bind_addrs(&self) -> impl ExactSizeIterator<Item = (&String, u16)> {
         self.core.binders.iter().map(|b| (&b.host, b.port))
     }
+
+    /// Add an rpc channel for an established connection represented as a file
+    /// descriptor.
```
BusyJay, line 566: Missing docs for `fd`.

BusyJay:
You need to signoff your commit to get DCO pass. The [Details](https://github.com/probot/dco#how-it-works) link has instructions about how to do it.

src/channel.rs
hcbdfinity: > You can mark `Channel::new` as public and unsafe instead.

sure, also fine with that solution -- changed

src/channel.rs
```diff
@@ -548,7 +570,11 @@ unsafe impl Send for Channel {}
 unsafe impl Sync for Channel {}
 
 impl Channel {
-    fn new(cq: CompletionQueue, env: Arc<Environment>, channel: *mut grpc_channel) -> Channel {
+    pub unsafe fn new(
```
BusyJay, line 573: Doc is needed.

BusyJay (commented):
Rest LGTM.

BusyJay (approved):
Thanks for your contributions! Please fix CI.

hcbdfinity:
> Thanks for your contributions! Please fix CI.

Thanks for review and sorry for slight delay fixing issues -- all green now, let me know if anything else is required to proceed.

hcbdfinity:
Gentle ping, anything that is required still for review?

BusyJay (approved):

BusyJay:
@hunterlxt PTAL

yiwu-arbug:
> Most of reviewers are on National holiday, I'm afraid it may not be merged till Friday.
> 
> /cc @yiwu-arbug can you take a look?

@BusyJay sorry, I didn't notice your comment :/

hunterlxt:
LGTM, but maybe it needs an unit test to verify that this api works

hicqu (approved):

BusyJay:
The method is only calling the unsafe one, I'm OK without a ut.

hunterlxt (approved):
LGTM

PR #492, opened 2020-10-26
*: update grpc c core to 1.33.1

BusyJay:
This PR updates grpc c core to 1.33.1. Upstream adds a new re2 module which makes the size of the crate too large. I optimize it by introducing a new `boringssl-src` crate. It can slow the compilation a little bit as it need to download extra package but the size of the library becomes 5.4MB now.


BusyJay:
Interesting, updating c core can bring 23% improvement in throughput in `rust_protobuf_async_unary_qps_unconstrained` case and 35% improvement in `rust_protobuf_async_streaming_qps_unconstrained`.

BusyJay:
I removed dynamic link check from arm CI as it's hard to maintain. And test code is also removed from bindgen as they are useless for users, they are still generated and tested in CI.

hunterlxt:
It seems that grpc does not support make to install

BusyJay:
I can't not reproduce the same error as travis arm reported when openssl feature is used on an arm laptop with the same openssl version. I remove the check for now as we support `openssl-vendored`, users have a workaround.

BusyJay:
@hunterlxt @hicqu PTAL

hunterlxt (approved):

hicqu (approved):

PR #495, opened 2020-11-11
codec: avoid overflowing error

BusyJay:
grpc c core can only handle message that is not larger than 4GiB.


src/codec.rs
```diff
@@ -31,16 +31,23 @@ pub mod pb_codec {
...
+        if cap <= u32::MAX as usize {
+            unsafe {
+                let bytes = buf.realloc(cap);
+                let raw_bytes = &mut *(bytes as *mut [std::mem::MaybeUninit<u8>] as *mut [u8]);
+                let mut s = CodedOutputStream::bytes(raw_bytes);
+                t.write_to_with_cached_sizes(&mut s).map_err(Into::into)
+            }
+        } else {
+            Err(Error::Codec(
+                format!("message is too large: {} > u32::MAX", cap).into(),
```
hunterlxt, line 49: Add comment to explain the limit is from somewhere in grpc core

src/codec.rs
```diff
@@ -56,20 +63,27 @@ pub mod pb_codec {
...
+            unsafe {
+                let bytes = buf.realloc(size);
+                let mut b = &mut *(bytes as *mut [std::mem::MaybeUninit<u8>] as *mut [u8]);
+                msg.encode(&mut b)?;
+                debug_assert!(b.is_empty());
+            }
+            Ok(())
+        } else {
+            Err(Error::Codec(
+                format!("message is too large: {} > u32::MAX", size).into(),
```
hunterlxt, line 85: same as above

hunterlxt (commented):
Most LGTM

src/codec.rs
hunterlxt: refer to https://github.com/tikv/grpc-rs/pull/496/files#diff-ca14eabd5ab83f72548cd7e44eaea3f56a5d2808e3275c6b2a97622ebe5d1dd0R9-R11 ?

JmPotato:
Any updates?

ekexium:
Any updates?

hunterlxt:
seem that CI should be fixed first

BusyJay:
The problem is it's not enough to only include this check as rust-protobuf doesn't return correct length.

BusyJay:
I have sent https://github.com/stepancheg/rust-protobuf/pull/571 to fix the panic. This PR can be merged first.

BusyJay:
Nightly failure is a known issue: https://github.com/tokio-rs/prost/issues/526.

hicqu (approved):

BusyJay:
The changed public APIs are all from codec. Those APIs are used by generated code, and those code can work with the new API without any changes. So I plan to land it without bumping minor versions to make it possible to fix overflowing error for previous versions.

Connor1996 (approved):
LGTM

PR #498, opened 2020-11-18
grpc-sys: Upgrade bindgen to the latest version

heftig:
As requested in https://github.com/tikv/grpc-rs/pull/452#issuecomment-729693080.


BusyJay:
Thanks @heftig!

BusyJay:
I tried to compile and test the project on the new Apple M1, this patch seemed required to get it compile.

grpc-sys/Cargo.toml
```diff
@@ -65,5 +65,5 @@ cmake = "0.1"
 pkg-config = "0.3"
 walkdir = "2.2.9"
 # Because of rust-lang/cargo#5237, bindgen should not be upgraded util a minor or major release.
-bindgen = { version = "0.51.0", default-features = false }
+bindgen = { version = "0.55.1", default-features = false, features = ["runtime"] }
```
BusyJay, line 68: How about updating it to 0.56?

grpc-sys/build.rs
```diff
@@ -298,6 +298,8 @@ fn bindgen_grpc(mut config: bindgen::Builder, file_path: &PathBuf) {
         .clang_arg("-std=c++11")
         .rustfmt_bindings(true)
         .impl_debug(true)
+        .size_t_is_usize(true)
+        .disable_header_comment()
```
BusyJay, line 302: I suggest to keep it so people may not try to edit the generated file by hand.

heftig: The header comment now contains the bindgen version number, making the output less stable.

BusyJay: Get it.

BusyJay (approved):

hunterlxt (approved):

PR #499, opened 2020-11-19
Add a way to not use bindgen

### LINKED ISSUE #489: Feature to conditionally include bindgen dependency
Hello 👋 

We are facing a problem similar than [this one](https://github.com/rust-lang/cargo/issues/5237), in our dependency tree we have multiple versions of `bindgen` and that fails to compile 😢 

Since you are already committing the bindings in tree, for the Linux on x86 and Aarch64 targets, `bindgen` is not needed for those targets, if the bindings do not need to be updated with the `UPDATE_BIND` env var.

Because of that, I believe that the `bindgen` dependency (and maybe more) could be removed from `grpc-sys` with a special feature. That would make compilation much faster and fix those `failed to select a version for clang-sys` issues!

hug-dev:
Fix #489

Add the `use-bindgen` feature which is activated by default.

If disabled, then the previously generated bindings will be used instead of generating new ones. It will fail compilation if this feature is disabled for the non supported targets.

If enabled, the behaviour is the same as before.


hug-dev:
hmm `--no-default-features` is used with targets that do not have pregenerated bindings... I can instead make a `not-use-bindgen` feature and reverse the logic everywhere?

BusyJay:
How about using https://github.com/rust-lang/cargo/issues/7914 to enable use-bindgen for unsupported targets?

hug-dev:
> How about using rust-lang/cargo#7914 to enable use-bindgen for unsupported targets?

That is a good idea but I think the `cfg` parameters [can not filter](https://doc.rust-lang.org/reference/conditional-compilation.html#set-configuration-options) on the target triple individually... So I don't think I can use that to enable `bindgen` on every target but Aarch64 and x86 on Linux.

I fixed the Mac CI workflow [here](https://github.com/tikv/grpc-rs/blob/4112a2249ebaa7a354cf59179979e3b4b08a077f/.github/workflows/ci.yml#L78) by enabling the `use-bindgen` feature there.

I am aware though that this will break compilation for everyone using this crate with `--no-default-features` on unsupported platforms.

BusyJay:
> ...the cfg parameters can not filter on the target triple individually...

How about:
```
[target.'cfg(not(all(any(target_arch="x86_64", target_arch="aarch64"), target_os="linux")))'.dependencies.grpcio-sys]
path = "grpc-sys"
version = "0.7"
features = ["use-bindgen"]
```
?

hug-dev:
ooh yes I guess that would do it, nice 👌 

With the following:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 70b1b39..a02033d 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ autoexamples = false
 all-features = true

 [dependencies]
-grpcio-sys = { path = "grpc-sys", version = "0.7", default-features = false }
+grpcio-sys = { path = "grpc-sys", version = "0.7" }
 libc = "0.2"
 futures = "0.3"
 protobuf = { version = "2.0", optional = true }
@@ -30,7 +30,7 @@ parking_lot = "0.11"
 members = ["proto", "benchmark", "compiler", "interop", "tests-and-examples"]

 [features]
-default = ["protobuf-codec", "secure", "use-bindgen"]
+default = ["protobuf-codec", "secure"]
 protobuf-codec = ["protobuf"]
 prost-codec = ["prost", "bytes"]
 secure = ["grpcio-sys/secure"]
@@ -47,3 +47,6 @@ travis-ci = { repository = "tikv/grpc-rs" }

 [patch.crates-io]
 grpcio-compiler = { path = "compiler", version = "0.7.0", default-features = false }
+
+[target.'cfg(not(all(any(target_arch="x86_64", target_arch="aarch64"), target_os="linux", target_env="gnu")))'.dependencies]
+grpcio-sys = { path = "grpc-sys", version = "0.7", features = ["use-bindgen"] }
diff --git a/grpc-sys/Cargo.toml b/grpc-sys/Cargo.toml
index 60e1d2f..c02f7eb 100644
--- a/grpc-sys/Cargo.toml
+++ b/grpc-sys/Cargo.toml
@@ -53,7 +53,6 @@ openssl-sys = { version = "0.9", optional = true, features = ["vendored"] }
 libz-sys = { version = "1.0.25", features = ["static"] }

 [features]
-default = ["use-bindgen"]
 secure = []
 openssl = ["secure"]
 openssl-vendored = ["openssl", "openssl-sys"]
```

compiling with `cargo build -Z features=itarget` with `nightly` rust, the `use-bindgen` feature is not enabled on my machine (x86 linux gnu). I can enable it adding `--features use-bindgen`.

I guess that would do it but the main problem I see is that this requires a `nightly` toolchain to work... For my use-case ideally `stable` would be used.

BusyJay (dismissed):

BusyJay:
Got it. I guess this can only be brought to 0.8.0.

hunterlxt (approved):

hug-dev:
Hello @BusyJay !

Do you know when `0.8.0` is planned to be released? We would be happy to use the change of this PR 😃

hug-dev:
Hello @BusyJay 

Kindly pinging you again to know about plans for a next release 👌

BusyJay:
Hi, @hug-dev, we are working on #506, when it's merged we may release next version.

hug-dev:
Excellent! Sorry for the spam 😋

hug-dev:
Thanks for the release 👌

PR #505, opened 2020-12-28
grpc-sys: Use grpc headers found by pkgconfig.

ssloboda:
Ran into this issue when cross compiling after updating libraries. I had originally fixed and tested this on v0.7.1 but it looks like a PR was just merged so I rebased. Here's the original fix in case you're interested.  https://github.com/StarryInternet/grpc-rs/tree/ssloboda/fix-cross-compile-bindgen-v0.7.1

Happy to refactor if needed.


ssloboda:
sorry, fixing the CI failures now

BusyJay:
Can you explain more about the error? The headers should be the same files across platform.

ssloboda:
From the commit message:
> Previously the bundled grpc headers were always used to generate bindings even in the case where pkgconfig was used to locate gRPC to link against. This caused errors at runtime, e.g., SEGFAULT, rather than a compile-time error. To fix this, use the headers found by pkgconfig to generate bindings rather than than the bundled headers if pkgconfig is
already being used.

This is not a problem that I expect to come up often however it will make things easier for those who dynamically link against a system installation of gRPC that may not match the bundled version exactly. This allows for semver compatible version bumps independent of the `grpcio` crate.

Does that clear things up?

grpc-sys/build.rs
```diff
@@ -307,7 +307,7 @@ fn bindgen_grpc(file_path: &PathBuf) {
     let cfg = config
         .header("grpc_wrap.cc")
         .clang_arg("-xc++")
-        .clang_arg("-I./grpc/include")
+        .clang_arg(format!("-I{}", grpc_include_dir.display()))
```
BusyJay, line 310: This doesn't seem to be correct. For example, include directory should be "/usr/local/include" instead of "/usr/local/include/grpc".

BusyJay:
> ...that may not match the bundled version exactly. 

Everytime updating the bundled version, there has to be some adaption here or there. So I doubt if it can work correctly. Though I agree compile time error is better than runtime.

grpc-sys/build.rs
ssloboda: :man_facepalming:

ssloboda:
Thanks for the review @BusyJay!

I've address the issues you pointed out in a fixup commit for ease of reviewing, will squash once you give the okay. Let me know if there's anything else you want changed.

grpc-sys/build.rs
```diff
@@ -400,15 +400,28 @@ fn main() {
...
     }
 
-    if get_env("GRPCIO_SYS_USE_PKG_CONFIG").map_or(false, |s| s == "1") {
+    let grpc_include_dir = if get_env("GRPCIO_SYS_USE_PKG_CONFIG").map_or(false, |s| s == "1") {
         // Print cargo metadata.
         let lib_core = probe_library(library, true);
+        let grpc_include_dir = lib_core
+            .include_paths
+            .iter()
+            .find(|grpc_inc_path| grpc_inc_path.is_dir())
```
BusyJay, line 409: Is it necessary to also check whether `grpc` subdirectory also exists?

BusyJay (commented):
Rest LGTM

BusyJay:
Hi @ssloboda do you have time to address the comment and sign off your commits? We may release next version soon, it will be nice to deliver your patch as well.

PR #506, opened 2021-01-28
update grpc to 1.35.0

BusyJay:


hunterlxt (commented):
Does grpc-sys need to be updated?

BusyJay:
Seems compiler takes too much memory that it's always killed on travis arm machine.

BusyJay:
Limiting the concurrent jobs on arch doesn't seem to meet travis's requirement. I have checked and tested it manually in an arm linux server, it works perfectly. I'm going to remove the arm linux required CI status. We may need to test it manually for arm before every release as a short time workaround.

.travis.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     arch: arm64
```
sticnarf, line 29:
```suggestion
    arch: arm64-graviton2
    vm: virt
```
Travis arm64 uses LXD by default. I guess that it causes the core number detected is quite large. In my experience, switching to use VM solves this problem. Please have a try.

BusyJay:
I tested it on a aws graviton vm and was unable to reproduce the failure.

hunterlxt (approved):

hicqu (approved):

PR #510, opened 2021-02-19
proto: add missing features

BusyJay:
otherwise it will not compile on hosts that have no pre-generated bindings.


sticnarf (approved):
LGTM

hunterlxt (approved):

PR #511, opened 2021-03-02
grpc-sys: detect failure early

BusyJay:
grpc c core 1.35 requires gcc 4.9+. It usually makes users switch
compiler on failure. But cmake doesn't work well on compiler changes, so
this pr detects the changes early and remove stale cache so that it can
finish compilation.


hunterlxt (dismissed):
LGTM

sticnarf (approved):

hunterlxt (approved):

BusyJay:
I believe rust-rocksdb also has the same problem. /cc @yiwu-arbug

PR #513, opened 2021-03-03
*: support connectivity related APIs

### LINKED ISSUE #483: Support connectivity state query
Connectivity state APIs described in https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md is very useful for connection life cycle maintenance.

To implement the docs, we need two APIs, first is `get_state` that returns the state of the channel, second is `wait_for_state_change` that returns a future represent if the state of the channel has changed.

BusyJay:
This PR adds the missing documented API `wait_for_state_change` to
channel, and add a helper function to wait for connected easily.

See also https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md.

Close #483.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -8,8 +8,9 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::{cmp, i32, ptr};
 
-use crate::grpc_sys::{
-    self, gpr_timespec, grpc_arg_pointer_vtable, grpc_channel, grpc_channel_args,
+use crate::{
+    grpc_sys::{self, gpr_timespec, grpc_arg_pointer_vtable, grpc_channel, grpc_channel_args},
+    Deadline,
 };
 use libc::{self, c_char, c_int};
 
@@ -17,6 +18,7 @@ use crate::call::{Call, Method};
 use crate::cq::CompletionQueue;
 use crate::env::Environment;
 use crate::error::Result;
+use crate::task::CallTag;
 use crate::task::Kicker;
 use crate::CallOption;
 use crate::ResourceQuota;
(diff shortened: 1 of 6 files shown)


src/channel.rs
```diff
@@ -589,12 +591,63 @@ impl Channel {
...
-    // changing the state.
+    /// If try_to_connect is true, the channel will try to establish a connection, potentially
+    /// changing the state.
     pub fn check_connectivity_state(&self, try_to_connect: bool) -> ConnectivityState {
         self.inner.check_connectivity_state(try_to_connect)
     }
 
+    /// Blocking wait for channel state change or deadline expiration.
+    ///
+    /// `check_connectivity_state` needs to called to get the current state.
```
hicqu, line 602: `s/to called/called/`?

hicqu:
rest LGTM.

hicqu (dismissed):

src/channel.rs
```diff
@@ -589,12 +591,63 @@ impl Channel {
...
+    /// If try_to_connect is true, the channel will try to establish a connection, potentially
+    /// changing the state.
     pub fn check_connectivity_state(&self, try_to_connect: bool) -> ConnectivityState {
         self.inner.check_connectivity_state(try_to_connect)
     }
 
+    /// Blocking wait for channel state change or deadline expiration.
+    ///
+    /// `check_connectivity_state` needs to be called to get the current state.
+    pub async fn wait_for_state_change(
```
hunterlxt, line 603: indicate return true means xxx

src/server.rs
```diff
@@ -523,14 +524,18 @@ pub fn request_call(ctx: RequestCallContext, cq: &CompletionQueue) {
 
 /// A `Future` that will resolve when shutdown completes.
 pub struct ShutdownFuture {
-    cq_f: CqFuture<()>,
+    cq_f: CqFuture<bool>,
```
hunterlxt, line 527: add comments to indicate what bool mean

tests-and-examples/tests/cases/misc.rs
```diff
@@ -273,3 +273,86 @@ impl ServerChecker for FlagChecker {
...
+        ch.check_connectivity_state(false),
+        ConnectivityState::GRPC_CHANNEL_IDLE
+    );
+
+    // It can't be ready since no server is running.
+    assert!(!block_on(ch.wait_for_connected(Duration::from_millis(100))));
+    assert!(block_on(ch.wait_for_state_change(
+        ConnectivityState::GRPC_CHANNEL_CONNECTING,
+        Duration::from_secs(3)
+    )));
```
hunterlxt, line 330: irrelevant question: If the server does not start, does it always keep this state?

BusyJay: It will keep trying reconnecting until timeout (with no pending requests), at that time, it becomes Idle.

hunterlxt (approved):
LGTM

PR #514, opened 2021-03-04
*: support rich error

### LINKED ISSUE #443: Support the "Richer error model"

BusyJay:
It's a common practice to return a protobuf struct as error status
instead of simple string. This PR adds complete rich error support
by following C++'s implementations. It changes several APIs to make
code clean and match other languages' design.

For more informations about rich error model, you can also take a
look at https://grpc.io/docs/guides/error/#richer-error-model.

Close #443.


src/error.rs
```diff
@@ -38,10 +38,19 @@ pub enum Error {
...
-            },
+            Error::RpcFailure(s) => {
+                if s.error_details().is_empty() && s.error_message().is_empty() {
+                    write!(fmt, "RpcFailure: {}", s.error_code())
+                } else {
+                    write!(
+                        fmt,
+                        "RpcFailure: [{}] binary: {} {}",
+                        s.error_code(),
+                        !s.error_details().is_empty(),
```
kennytm, line 49: that `binary: true/false` part is quite obscure to people reading the error message

BusyJay: Any suggestions?

kennytm: I think we should simply ignore it (keep the old `Display` output). Saying that `.error_details()` exists mean nothing in the log since you can't retrieve the details again, while printing the entire `.error_details()` into the log is too noisy.

kennytm (commented):
LGTM.

gregwebs:
Will this PR be updated and merged? Error information is fundamental for a system to be able to operate well.

BusyJay:
Yes, but I'm focus on releasing 5.0 right now and the PR needs another LGTM.

proto/src/util.rs
```diff
@@ -36,3 +38,53 @@ pub fn create_test_channel_credentials() -> ChannelCredentials {
...
+    type Error = grpcio::Error;
+
+    fn try_from(value: grpcio::RpcStatus) -> grpcio::Result<Self> {
+        let mut s = Status::default();
+        #[cfg(feature = "protobuf-codec")]
+        protobuf::Message::merge_from_bytes(&mut s, value.error_details())?;
+        #[cfg(feature = "prost-codec")]
+        prost::Message::merge(&mut s, value.error_details())?;
+        if s.code == value.error_code().into() {
+            if s.message == value.error_message() {
```
gregwebs, line 52: This seems like a temporary check that doesn't need to stay in when we are confident the new version is working?

src/call/mod.rs
```diff
@@ -156,10 +156,15 @@ impl<Req, Resp> Method<Req, Resp> {
 #[derive(Debug, Clone)]
 pub struct RpcStatus {
     /// gRPC status code. `Ok` indicates success, all other values indicate an error.
-    pub status: RpcStatusCode,
+    code: RpcStatusCode,
```
gregwebs, line 159: Why change the field name?

src/call/mod.rs
```diff
@@ -170,16 +175,49 @@ impl Display for RpcStatus {
...
 
     /// Create a new [`RpcStatus`] that status code is Ok.
     pub fn ok() -> RpcStatus {
-        RpcStatus::new(RpcStatusCode::OK, None)
+        RpcStatus::new(RpcStatusCode::OK)
+    }
+
+    /// Return the instance's error code.
+    #[inline]
+    pub fn error_code(&self) -> RpcStatusCode {
```
gregwebs, line 207: It seems that this is not necessarily an error code in that the code could be OK.

src/error.rs
```diff
@@ -38,10 +38,18 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Error::RpcFailure(RpcStatus { status, details }) => match details {
-                Some(details) => write!(fmt, "RpcFailure: {} {}", status, details),
-                None => write!(fmt, "RpcFailure: {}", status),
-            },
+            Error::RpcFailure(s) => {
+                if s.error_message().is_empty() {
+                    write!(fmt, "RpcFailure: {}", s.error_code())
```
gregwebs, line 43: I think it would help to have the same format whether there is an error message or not. I would tend towards less punctuation: `RpcFailure {}: {}, s.error_code(), s.error_message()`

tests-and-examples/tests/cases/metadata.rs
```diff
@@ -23,6 +25,20 @@ impl Greeter for GreeterService {
             self.tx.send((key.to_owned(), value.to_owned())).unwrap();
         }
 
+        if req.name == "root" {
+            let mut status = Status::default();
```
gregwebs, line 29: Better to use `Status::with_message` in examples? I am wondering what the use case is for generating a default when these constructors are available.

gregwebs (dismissed):
Just some minor questions. LGTM

proto/src/util.rs
BusyJay: The library doesn't know whether the other end, which can be other library or even in different language, is implemented correctly.

src/call/mod.rs
BusyJay: Because I want to name the new field as `details` just like other implementations. It's going to be a breaking change, so renaming all fields to keep consistent with protobuf struct for better understanding.

src/call/mod.rs
BusyJay: Indeed, I just follow the name of the C++'s version.

src/error.rs
BusyJay: It will leave some blank spaces in the end of string.

tests-and-examples/tests/cases/metadata.rs
BusyJay: `Status` here is from protobuf, which doesn't have `with_message` constructor, instead of the one in grpcio. And using `default` is to get it compile for both prost and protobuf.

src/error.rs
gregwebs: I don't know if that is a problem. But mostly I don't see the point of putting the code in brackets.

proto/src/util.rs
gregwebs: Still I am wondering if in production this is the best behavior? It seems like something where I would prefer to log a warning but return Ok since the code is still correct.

BusyJay: Then there is no way to detect the error. If user want to ignore the error, they can still parse the status from `error_details` directly.

BusyJay:
@gregwebs PTAL

gregwebs (approved):
LGTM

PR #516, opened 2021-03-09
channel: fix send issue in state change API

BusyJay:
async/await doesn't work well with temporary variables, this PR changes
it to `impl Future` to get around Send requirement.


hunterlxt (approved):

Connor1996 (approved):
LGTM

PR #518, opened 2021-03-12
*: implements health check

BusyJay:
This PR provides simple health check implementations referring to the go
version. It provides a `HealthService` to maintain the service statuses
and wake up watchers.

I use a standalone crate to avoid introduce grpc specific code into the grpcio
crate. And I don't reuse grpcio-proto to avoid the dependency of protobuf-build
which just make things complicated by introducing a lot of dependencies and
resulting in a dependency circle.


BusyJay:
Most code are generated by protobuf and prost, you can just skip them as long as it compiles.

health/src/service.rs
```diff
@@ -0,0 +1,237 @@
...
+        StatusCast {
+            state: AtomicUsize::new(VERSION_STEP | (status as usize)),
+            subscribers: Mutex::default(),
+        }
+    }
+
+    /// Updates the status to specified one and update version.
+    fn broadcast(&self, status: ServingStatus) {
+        let mut subscribers = self.subscribers.lock().unwrap();
+        let state = self.state.load(Ordering::Acquire);
```
sticnarf, line 42: Guarded by the `subscribers` mutex, I think it can be `Relaxed`.

health/src/service.rs
```diff
@@ -0,0 +1,237 @@
...
+        let cur_state = s.cast.state.load(Ordering::Relaxed);
+        if cur_state != s.last_state {
+            let status = state_to_status(cur_state);
+            s.last_state = cur_state;
+            return Poll::Ready(Some(status));
+        }
+
+        let mut subscribers = s.cast.subscribers.lock().unwrap();
+
+        let cur_state = s.cast.state.load(Ordering::Acquire);
```
sticnarf, line 83: It can be `Relaxed` because a mutex has memory fence itself.

health/src/service.rs
```diff
@@ -0,0 +1,237 @@
...
+            subscribers: Mutex::default(),
+        }
+    }
+
+    /// Updates the status to specified one and update version.
+    fn broadcast(&self, status: ServingStatus) {
+        let mut subscribers = self.subscribers.lock().unwrap();
+        let state = self.state.load(Ordering::Acquire);
+        let new_state = ((state + VERSION_STEP) & !STATUS_MASK) | (status as usize);
+        self.state.store(new_state, Ordering::Release);
```
sticnarf, line 44: And because L74 loads with `Relaxed`, this can be `Relaxed` as well.

BusyJay:
I add back protobuf-build to make it work well with TiKV. TiKV seems to use custom protobuf patch which doesn't work well with latest published one.

sticnarf (dismissed):
LGTM

Connor1996 (commented):
rest LGTM

health/src/service.rs
```diff
@@ -0,0 +1,237 @@
...
+/// is any updates.
+struct StatusCast {
+    state: AtomicUsize,
+    subscribers: Mutex<HashMap<u64, Waker>>,
+}
+
+impl StatusCast {
+    fn new(status: ServingStatus) -> StatusCast {
+        StatusCast {
+            state: AtomicUsize::new(VERSION_STEP | (status as usize)),
```
Connor1996, line 34: why we can't use status directly?

health/src/service.rs
```diff
@@ -0,0 +1,237 @@
...
+        let cur_state = s.cast.state.load(Ordering::Relaxed);
+        if cur_state != s.last_state {
+            let status = state_to_status(cur_state);
+            s.last_state = cur_state;
+            return Poll::Ready(Some(status));
+        }
+
+        let mut subscribers = s.cast.subscribers.lock().unwrap();
+
+        let cur_state = s.cast.state.load(Ordering::Relaxed);
```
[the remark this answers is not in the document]
BusyJay: The first check is a fast path. Only section protected by lock can guarantee `state` is not updated by others.

health/src/service.rs
BusyJay: Then you can't solve the ABA problem.

Connor1996 (dismissed):
LGTM

PR #521, opened 2021-03-31
*: provide default service implementations

BusyJay:
According to grpc interop [test descriptions][1], calling unimplemented method should return
UNIMPLEMENTED status code. Before this PR, user is required to implement all methods
and return UNIMPLEMENTED explicitly if desired. This PR changes the behavior by providing
a default implementation, so that it can truely meet the description.

[1]: https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#unimplemented_method


hunterlxt:
LGTM but little suggestion, provide simple comment to describe when to use `unimplemented_call!`

hunterlxt (approved):

hunterlxt (approved):

hicqu (approved):

PR #522, opened 2021-04-06
*: use xtask workflow

BusyJay:
This PR removes all scripts for maintaining the project. Instead,
introduce [xtask][1] workflow to whole process more portable and rusty.

It also introduces a new workflow named codegen to generate code for
health crate.

[1]: https://github.com/matklad/cargo-xtask


xtask/src/main.rs
```diff
@@ -0,0 +1,282 @@
...
+    cmd
+}
+
+fn submodule() {
+    exec(cmd("git").args(&["submodule", "update", "--init", "grpc-sys/grpc"]));
+    for dir in &["cares/cares", "abseil-cpp", "re2"] {
+        exec(cmd_in("git", "grpc-sys/grpc/third_party").args(&[
+            "submodule",
+            "update",
+            "--init",
```
hicqu, line 114: Should add `--recursive`? I can't compile without the flag if I haven't done it.

hicqu:
It's necessary to make it can be compiled with rustc nightly 2021-01-16, or update toolchain for TiKV.
Rest LGTM!

hicqu (approved):

hicqu (approved):

BusyJay:
xtask is only used for development, so it doesn't matter even if rustc used in TiKV can't compile it.

xtask/src/main.rs
BusyJay: It's a for loop executed in the directory, so `recursive` is not necessary. On the other hand, recursive will also clone other unnecessary dependencies.

hunterlxt (commented):
LGTM

PR #530, opened 2021-05-26
*: make 0.7.x work with 0.9.x compiler

### LINKED ISSUE #528: `grpcio-proto` 0.7 and 0.8 now fail to build as a dependency
**Describe the bug**
`grpcio-proto` 0.8 brings in `protobuf-build`. `protobuf-build` versions 12.1+  depend on `grpcio-compiler>=0.8.0`, which as of today brings in `grpcio-compiler` 0.9.0, which causes the build to fail when building `grpc-proto` 0.8 as a dependency.

`grpcio-proto` 0.7 (and maybe older versions) has the same issue, because `protobuf-build` 0.11.4 brings in `grpcio-compiler>=0.5.0`.

This can be worked around with cargo patches, but it is a bit painful to do so.

**To Reproduce**
1. Make a new crate and add `grpcio-proto = "0.8"` to the dependencies array in its `Cargo.toml`.
2. `cargo check`

```
    Checking grpcio-proto v0.8.1
error[E0433]: failed to resolve: could not find `unimplemented_call` in `grpcio`
   --> /home/matthew/misc/grpc-test/target/debug/build/grpcio-proto-ca27f1d4f4011d16/out/testing/test_grpc.rs:189:17
    |
189 |         grpcio::unimplemented_call!(ctx, sink)
    |                 ^^^^^^^^^^^^^^^^^^ could not find `unimplemented_call` in `grpcio`
<many copies of the same error...>
```

**Expected behavior**
Old versions of `grpcio-protobuf` should continue to build as dependencies without cargo-patching inner dependencies when a new version is released. Unfortunately, I think this can only be done by yanking `grpcio-compiler` 0.9.0 and/or releasing a version of it that works with old `grpcio-proto` versions.

Alternatively, new patch releases of `protobuf-build` and/or old versions of `grpcio-proto` to fix the issue might help, but I haven't fully thought through how that would need to work -- I understand there's a cyclic dependency the `>=` dependency fixes (at the cost of breaking old versions' builds).

**System information**
* CPU architecture: x86_64
* Distribution and kernel version: Fedora 33, 5.11.18-200.fc33.x86_64
* SELinux on?: I think so
* Any other system details we should know?:

**Additional context**

BusyJay:
Close #528.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.7.2 - 2021-05-28
+
+- Work with 0.9.x compiler
+
 # 0.7.1 - 2020-12-18
 
 - Allow CXX environment variable to override g++ for musl build (#500)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.7.1"
+version = "0.7.2"
 edition = "2018"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
--- a/compiler/src/util.rs
+++ b/compiler/src/util.rs
@@ -34,7 +34,7 @@ impl<'a> Iterator for NameSpliter<'a> {
         let mut meet_lower = false;
         for i in self.pos..self.name.len() {
             let c = self.name[i];
-            if b'A' <= c && c <= b'Z' {
+            if (b'A'..=b'Z').contains(&c) {
                 if meet_lower {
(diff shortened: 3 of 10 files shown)


sticnarf (approved):
LGTM

hunterlxt (approved):

BusyJay:
Remaining failures are caused by oudated test certificates and outdated absl library. Going to merged without extra fixes.

Issue #460, opened 2020-04-29
Support sending initial metadata and trailing metadata


BusyJay:
Currently, only sending initial metadata from client side and receiving initial metadata from server side are supported. The missing parts are:

- [ ] Sending initial metadata from server side
- [ ] Receiving initial metadata from client side
- [ ] Sending/Receiving trailing metadata

BusyJay:
After investigating the design of other languages, I propose the following API designs:

### Sending metadata from client

This is supported already in `CallOption`.

```
let builder = MetadataBuilder::new();
let opt = CallOption::default().headers(builder.build());
```

### Receiving metadata from client

Introducing a new struct is a way, but it adds extra maintenance. We can add extra methods for two Receivers instead.
```
pub async fn headers(&mut self) -> Metadata {...}
pub async fn trailer(&mut self) -> Metadata {...}
```
So only one struct needs to be used and if people are not interested in metadata, they don't need to change their code.

### Sending metadata from server

Similar to receiving metadata from client, we can also use separate methods for server sink.
```
pub async fn send_headers(&mut self, header: Metadata);
pub fn set_headers(&mut self, header: Metadata);
pub fn set_trailer(&mut self, trailer: Metadata);
```

`set_headers` and `set_trailer` set the metadata and sends lazily. `send_headers` sets the header and send immediately.

### Receiving metadata from server

This is supported already.

```
for (key, val) in ctx.request_headers() {
    ...
}
```

The interface design are mainly copied from [golang](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md), so this can also be friendly for users with similar background. One thing need to take care is that we need extra validation checks like headers set after being sent is meaningless.

What do you think? /cc @christian-oudard

christian-oudard:
@BusyJay
For the use case at MobileCoin, we will need streaming metadata as well. I will start implementing this API in our pull request, including streaming support.

christian-oudard:
@BusyJay, please take a look at the code I have written so far in PR #543.

I have some questions about how to continue the implementation.

First off, I was unable to make the API spec you gave me work. The problem is that implementing `Future` on `ClientUnaryReceiver` means that `.await` does a *move* of the receiver into the await call, which does not allow us to do anything else afterwards, such as getting metadata. I solved this by adding an async function, `.message()`, which waits for the `BatchFuture` on a borrowed `self`.

[...]

PR #532, opened 2021-06-08
Client-side unary-call metadata support.

christian-oudard:
* Add methods to BatchContext to get metadata.
* Modify BatchFuture to read/return those as well.
* Update existing code paths to discard metadata read by BatchFuture.
* Create a new FullClientUnaryReceiver structure to explicitly pass it along.
* Add async/sync accessors to those details.
* Add unary metadata to client codegen.

This was written by my colleague James Cape last year; we are just now getting around to sending a pull request.


BusyJay:
Interesting. Sending metadata is supported already using `CallOption`. For receiving I need some time to think about the interface design. Will get back soon.

BusyJay:
Let's talk about it in #460.

christian-oudard:
Reimplementing with new API in https://github.com/tikv/grpc-rs/pull/543

PR #537, opened 2021-06-22
Make boringssl-src optional

### LINKED ISSUE #536: Remove dependency on boringssl-src if secure is not activated
Hello 👋 

In the README, it is [written](https://github.com/tikv/grpc-rs#feature-secure):

> secure feature enables support for TLS encryption and some authentication mechanism. When you do not need it, for example when working in intranet, you can disable it by using the following configuration

However, even if it is disabled, `boringssl-src` will be included in the dependencies via `grpcio-sys` (see [here](https://github.com/tikv/grpc-rs/blob/master/grpc-sys/Cargo.toml)).

Removing the need of compiling `boringssl-src` when the `secure` feature is disabled will make compilation faster and also easier for cross-compilation targets 👍 

I am happy to have a go at it and try to modify the `build.rs` file to something that'd work.

hug-dev:
If the secure feature is disabled, it's not needed to compile the boringssl-src crate.

This assume that the `secure` feature is using `boringssl-src` by default which I think is the current behaviour.

A bad aspect of this is that `boringssl-src` will be compiled even if `openssl` is chosen but I think that is currently the case anyway?

Fix #536

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,8 @@ on:
 
 env:
   RUST_BACKTRACE: 1
-  RUSTFLAGS: "--deny=warnings"
+  # Some of the bindgen tests generate "deref-nullptr" warnings, see https://github.com/rust-lang/rust-bindgen/issues/1651
+  RUSTFLAGS: "--deny=warnings --allow deref-nullptr"
   TEST_BIND: 1
 
 jobs:
@@ -100,6 +101,7 @@ jobs:
     runs-on: windows-latest
     env:
       LIBCLANG_PATH: 'C:\Program Files\LLVM\bin'
+      RUSTFLAGS: ""
     steps:
     - uses: actions/checkout@v2
     - run: choco install -y llvm
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -172,7 +172,11 @@ fn build_grpc(cc: &mut cc::Build, library: &str) {
         config.define("gRPC_BUILD_CODEGEN", "false");
         // We don't need to build benchmarks.
         config.define("gRPC_BENCHMARK_PROVIDER", "none");
-        config.define("gRPC_SSL_PROVIDER", "package");
+
+        if cfg!(feature = "secure") {
+            config.define("gRPC_SSL_PROVIDER", "package");
+        }
+        #[cfg(feature = "secure")]
         if cfg!(feature = "openssl") {
             if cfg!(feature = "openssl-vendored") {
                 config.register_dep("openssl");
@@ -254,6 +258,7 @@ fn figure_ssl_path(build_dir: &str) {
     println!("cargo:rustc-link-lib=crypto");
 }
 
+#[cfg(feature = "secure")]
 fn build_boringssl(config: &mut CmakeConfig) {
     let boringssl_artifact = boringssl_src::Build::new().build();
     config.define(
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -173,6 +173,8 @@ fn build_grpc(cc: &mut cc::Build, library: &str) {
         // We don't need to build benchmarks.
         config.define("gRPC_BENCHMARK_PROVIDER", "none");
 
+        // `package` should only be set for secure feature, otherwise cmake will always search for
+        // ssl library.
         if cfg!(feature = "secure") {
             config.define("gRPC_SSL_PROVIDER", "package");
         }
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -395,6 +395,7 @@ impl ChannelBuilder {
 
     /// Build `ChannelArgs` from the current configuration.
     #[allow(clippy::useless_conversion)]
+    #[allow(clippy::cmp_owned)]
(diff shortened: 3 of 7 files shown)


BusyJay:
> A bad aspect of this is that boringssl-src will be compiled even if openssl is chosen but I think that is currently the case anyway?

Yes, it's expected. The optimal way is depend on boringssl-src if openssl feature is not enabled. But there seems to be no way to do so in Cargo.

hug-dev:
Seems like on Mac OpenSSL is needed for grpc compilation even if `secure` feature is not on? Do you have any idea of what went wrong 😬?

BusyJay:
> Seems like on Mac OpenSSL is needed for grpc compilation even if `secure` feature is not on? Do you have any idea of what went wrong 😬?

`gRPC_SSL_PROVIDER` should be set to `package` only when `secure` feature is configured.

hug-dev:
Thanks for the help! After *many* force pushes, all tests are passing 🌟

.github/workflows/ci.yml
```diff
@@ -43,7 +43,8 @@ jobs:
     - run: cargo build --no-default-features --features prost-codec
     - run: cd proto && cargo build --no-default-features --features prost-codec
     - run: cargo build
-    - run: cargo test --all
+      # Some of the bindgen tests generate warnings, see https://github.com/rust-lang/rust-bindgen/issues/1651
+    - run: RUSTFLAGS="" cargo test --all
```
BusyJay, line 47: How about adding `-A deref_nullptr` to `RUSTFLAGS`? So that other warnings can still be checked.

src/channel.rs
```diff
@@ -404,7 +405,7 @@ impl ChannelBuilder {
                     // On most modern compiler and architect, c_int is the same as i32,
                     // panic directly to simplify signature.
                     assert!(
-                        val <= i32::from(libc::INT_MAX) && val >= i32::from(libc::INT_MIN),
+                        val <= libc::INT_MAX && val >= libc::INT_MIN,
```
BusyJay, line 408: I think it's intended to convert as explained at L405.

BusyJay (commented):
You may also need to change .travis.yml

.github/workflows/ci.yml
hug-dev: Ah yes good point, will do that 👌

src/channel.rs
hug-dev: I see. On some platforms they might be different and that will error out. Will just allow the clippy lint there then.

.github/workflows/ci.yml
```diff
@@ -100,6 +101,7 @@ jobs:
     runs-on: windows-latest
     env:
       LIBCLANG_PATH: 'C:\Program Files\LLVM\bin'
+      RUSTFLAGS: ""
```
BusyJay, line 104: Why?

BusyJay (commented):
Rest LGTM

.github/workflows/ci.yml
hug-dev: Was about to comment about that :D With nothing there, there was an error that `-A deref-nullptr` was not recognised by `rustc`, with only `--deny=warnings` the errors were still showing. Not sure why

BusyJay: I guess maybe l102 overwrites L11.

hug-dev: Yes, I thought that as well, that putting an empty string here would empty the `RUSTFLAGS` defined before. It still failed but that seems to be for another reason?

BusyJay (dismissed):

BusyJay:
@hunterlxt PTAL

grpc-sys/build.rs
```diff
@@ -172,7 +172,11 @@ fn build_grpc(cc: &mut cc::Build, library: &str) {
         config.define("gRPC_BUILD_CODEGEN", "false");
         // We don't need to build benchmarks.
         config.define("gRPC_BENCHMARK_PROVIDER", "none");
-        config.define("gRPC_SSL_PROVIDER", "package");
+
+        if cfg!(feature = "secure") {
+            config.define("gRPC_SSL_PROVIDER", "package");
+        }
```
hunterlxt, line 178: why not just define the cmake config regardless of the `secure` feature

hunterlxt (approved):

grpc-sys/build.rs
BusyJay: @hug-dev can you add a comment here?

hug-dev: I did that to fix a failing test on Mac OS, after [your recommendation](https://github.com/tikv/grpc-rs/pull/537#issuecomment-866605705).
Would the following comment be ok?

```
// OpenSSL is needed for gRPC compilation when the secure feature is on.
```

BusyJay: How about
```
// `package` should only be set for secure feature, otherwise cmake will always search for ssl library.
```

hug-dev: Done!

BusyJay (approved):
Thanks!

hunterlxt (approved):
LGTM

PR #539, opened 2021-06-25
grpcio-health don't need grpcio with secure enabled

JakubOnderka:
Using `grpcio-health` in project that don't need secure grpc will just slow down compilation.


BusyJay:
The commit needs to be signed off.

PR #541, opened 2021-06-30
Add target bindgen in Cargo.toml

devillove084:
Add this for mac and others, can help when use `default-features = false, features = ["protobuf-codec"]` in the common Cargo.toml file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,3 +54,8 @@ travis-ci = { repository = "tikv/grpc-rs" }
 
 [patch.crates-io]
 grpcio-compiler = { path = "compiler", version = "0.9.0", default-features = false }
+
+[target.'cfg(any(not(linux), not(any(target_arch = "x86_64", target_arch = "aarch64"))))'.dependencies.grpcio]
+version = "0.9.0"
+default-features = false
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -54,3 +54,8 @@ travis-ci = { repository = "tikv/grpc-rs" }
 
 [patch.crates-io]
 grpcio-compiler = { path = "compiler", version = "0.9.0", default-features = false }
+
+[target.'cfg(any(not(linux), not(any(target_arch = "x86_64", target_arch = "aarch64"))))'.dependencies.grpcio]
```
BusyJay, line 58: It should be `grpcio-sys`.

Cargo.toml
```diff
@@ -54,3 +54,8 @@ travis-ci = { repository = "tikv/grpc-rs" }
...
+version = "0.9.0"
+default-features = false
+features = ["secure", "use-bindgen"]
```
BusyJay, line 61: "secure" is unnecessary.

BusyJay (commented):
"use-bindgen" in default list can be removed too.

PR #543, opened 2021-08-03
Server-side metadata for all call types

christian-oudard:
**Related features:**
See issue #460

**Description of feature:**
Implement metadata sent from the server and read by the client.

**Implementation:**
ClientUnaryReceiver gains a .headers() and .trailer() method to convey metadata sent by server.

**Checklist:**

* [x] `cargo fmt -- --check` passes.
* [x] `cargo +nightly clippy` has no warnings.
* [x] `cargo test` passes.


BusyJay:
Thanks! You need to get all your commits signed off to pass the DOC check.

src/metadata.rs
```diff
@@ -254,6 +254,15 @@ impl Drop for Metadata {
     }
 }
 
+impl From<*const grpc_metadata_array> for Metadata {
```
BusyJay, line 257: Instead of implementing `From` trait, you can implement a new c function to take the metadata. There is already a function `grpcwrap_metadata_array_move`, so `grpcwrap_batch_context_recv_initial_metadata` can be implemented as
```
GPR_EXPORT void GPR_CALLTYPE
grpcwrap_batch_context_take_initial_metadata(
    const grpcwrap_batch_context* ctx, grpc_metadata_array* initial_metadata) {
    grpcwrap_metadata_array_move(initial_metadata, &(ctx->send_initial_metadata));
}
```

And in rust side, just declare a new metadata and pass it to c:
```
let mut metadata = MetadataBuilder::new().build();
unsafe {
    grpcwrap_batch_context_take_initial_metadata(ctx, &mut metadata);
}
```

. So that we don't expose unsafe implementation to safe type.

christian-oudard:
@BusyJay, I implemented your suggestion, with a couple of snags. First, the function to take the metadata out of the context requires a reference to the underlying C struct, not to the rust Metadata struct. Is there a way to do this without making it a public tuple member how I did?

Also, the client segfaults when freeing memory at the end of program execution. I think that context is freeing a null pointer. Can you help debug this? You can reproduce the problem with the test code here: https://github.com/christian-oudard/rust_tutorial/tree/master/grpc_tutorial

BusyJay:
> ...without making it a public tuple member

It's defined as C representation, so it's safe to cast `*mut Metadata` to `*mut grpc_metadata_array`.

> the client segfaults when freeing memory at the end of program execution

You can run it with gdb, it will pause when core. It should provide helpful stacktrace for what's going wrong.

christian-oudard:
> It's defined as C representation, so it's safe to cast `*mut Metadata` to `*mut grpc_metadata_array`.
> 
> >

There does not appear to be any way to move out of the C struct without segfaulting, because of the use of Unref in the grpc C code. Please correct me if I'm wrong about this. I just implemented it with cloning the metadata instead of attempting to move it.

BusyJay:
> because of the use of Unref in the grpc C code.

Where did it called? Last time I checked, all it needs was just a pointer to an initialized struct.

In my assumption, it should be written like following:

```
let mut metadata = MetadataBuilder::new().build();
unsafe {
    grpcwrap_batch_context_take_initial_metadata(ctx, &mut metadata as *mut _ as *mut grpc_metadata_array);
}
```

christian-oudard:
Alright, that's server-side metadata implemented for all four call types. @BusyJay, Can you re-review?

grpc-sys/grpc_wrap.cc
```diff
@@ -641,20 +641,27 @@ GPR_EXPORT grpc_call_error GPR_CALLTYPE grpcwrap_call_recv_initial_metadata(
...
-  ops[1].op = GRPC_OP_SEND_INITIAL_METADATA;
-  ops[1].flags = 0;
-  ops[1].reserved = nullptr;
 
+  if (initial_metadata) {
+    ops[nops].op = GRPC_OP_SEND_INITIAL_METADATA;
+    grpcwrap_metadata_array_move(&(ctx->send_initial_metadata), initial_metadata);
+    ops[nops].data.send_initial_metadata.count = ctx->send_initial_metadata.count;
+    ops[nops].data.send_initial_metadata.metadata = ctx->send_initial_metadata.metadata;
+    ops[nops].flags = 0;
```
BusyJay, line 661: How about let caller to customize the value?

grpc-sys/grpc_wrap.cc
```diff
@@ -673,14 +680,17 @@ grpcwrap_call_send_close_from_client(grpc_call* call, void* tag) {
 GPR_EXPORT grpc_call_error GPR_CALLTYPE grpcwrap_call_send_status_from_server(
     grpc_call* call, grpcwrap_batch_context* ctx, grpc_status_code status_code,
     const char* status_details, size_t status_details_len,
-    grpc_metadata_array* trailing_metadata, int32_t send_empty_initial_metadata,
-    grpc_slice* optional_send_buffer, uint32_t write_flags, void* tag) {
+    grpc_metadata_array* initial_metadata, grpc_metadata_array* trailing_metadata,
+    int32_t send_empty_initial_metadata, grpc_slice* optional_send_buffer,
```
BusyJay, line 684: I suggest to remove `send_empty_initial_metadata`. Caller can pass an empty initial metadata if necessary.

grpc-sys/grpc_wrap.cc
christian-oudard: I don't understand. Customize which value? `initial_metadata` can be full, empty, or null. If it is null, it is not sent.

grpc-sys/grpc_wrap.cc
christian-oudard: Good idea, I'll do that.

grpc-sys/grpc_wrap.cc
BusyJay: I mean `flags`. There are several flags can be set for metadata.

https://github.com/pingcap/grpc/blob/rs-release/include/grpc/impl/codegen/grpc_types.h#L493-L504

christian-oudard: Thanks for clarifying. I'll add `initial_metadata_flags`.

BusyJay: call flags should be part of `CallOptions`, just like how header is pass.

christian-oudard: Okay, I did this by adding `call_flags` to `StreamingCallSink`, so that metadata on client-streaming calls can use these flags.

src/call/client.rs
```diff
@@ -242,18 +248,36 @@ impl<T> ClientUnaryReceiver<T> {
     }
 
     #[inline]
-    pub fn resp_de(&self, reader: MessageReader) -> Result<T> {
+    pub fn resp_de(&mut self, reader: MessageReader) -> Result<T> {
```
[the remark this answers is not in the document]
christian-oudard: Whoops. Fixed.

src/call/client.rs
```diff
@@ -242,18 +248,36 @@ impl<T> ClientUnaryReceiver<T> {
...
 
-impl<T> Future for ClientUnaryReceiver<T> {
-    type Output = Result<T>;
+    async fn wait_for_batch_future(&mut self) -> Result<()> {
+        if self.finished {
+            return Ok(());
+        }
+
+        let data = Pin::new(&mut self.resp_f).await?;
+        self.initial_metadata = data.initial_metadata.clone();
```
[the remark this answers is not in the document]
christian-oudard: Yes, done.

src/metadata.rs
```diff
@@ -224,6 +225,13 @@ impl Metadata {
         })
     }
 
+    pub fn from_raw(src: *const grpc_metadata_array) -> Self {
```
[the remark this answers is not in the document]
christian-oudard: I'm confused about how to do this. I tried it this way, and it gave me a segmentation fault:
`            let m = src as *const metadata;
            return metadata::from_raw_parts((*m).0.metadata, (*m).0.count, (*m).0.capacity);
`
Can you give an example of how this could work without cloning?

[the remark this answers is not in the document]
BusyJay: Like this:
```
let mut res = Metadata::with_capacity(0);
grpcio_sys::grpcwrap_batch_context_move_recv_initial_metadata(ctx, &mut res as *mut _);
```

The implementation of `grpcwrap_batch_context_move_recv_initial_metadata` is:
```
GPR_EXPORT void GPR_CALLTYPE
grpcwrap_batch_context_take_recv_initial_metadata(
    const grpcwrap_batch_context* ctx, grpc_metadata_array* res) {
    grpcwrap_metadata_array_move(res, &(ctx->recv_initial_metadata));
}
```

The key point is use a pointer, which point to the rust stack to receive the moved content from C side. What you write will leave two pointer points to the same metadata, the first one is src->metadata, the second is the returned value. So it will double free eventually. `grpcwrap_metadata_array_move` will handle the pointers correctly.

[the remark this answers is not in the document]
christian-oudard: I attempted to implement this, but it is still segfaulting. Can you take a look at the head of my branch and see if you can fix the segfault?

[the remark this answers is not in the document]
BusyJay: Can you also provide the source code?

christian-oudard:
One more question about the design of the new headers API:
Our usage of gRPC in the MobileCoin codebase relies exclusively on unary calls, with no asynchrony. Can you suggest an API to return headers without requiring users to convert to async?

The previous fork of grpc-rs we were using had a syntax like this:
```let (headers, hello_response, trailer) = client.hello_full(hello_request)```

This required codegen changes to add a "_full" method on generated clients. Should we re-implement this method, or do you have a suggestion for an improvement to this API?

**Comment by sticnarf** [comment]
```rust
let (header, message, trailer) = block_on(async {
    (
        receiver.headers().await.ok().cloned(),
        receiver.message().await.expect("no message"),
        receiver.trailer().await.ok().cloned(),
    )
});
```

Maybe this can be provided by `ClientUnaryReceiver` for convenience.

christian-oudard:
I implemented @sticnarf 's suggestion, creating a method on `ClientUnaryReceiver` called `receive_sync`. Thoughts?

src/call/client.rs
```diff
@@ -427,16 +504,19 @@ struct ResponseStreamImpl<H, T> {
     read_done: bool,
     finished: bool,
     resp_de: DeserializeFn<T>,
+    headers_f: BatchFuture,
```
BusyJay, line 507: When will it be used?

grpc-sys/grpc_wrap.cc
```diff
@@ -279,10 +279,19 @@ grpcwrap_request_call_context_destroy(grpcwrap_request_call_context* ctx) {
   gpr_free(ctx);
 }
 
-GPR_EXPORT const grpc_metadata_array* GPR_CALLTYPE
-grpcwrap_batch_context_recv_initial_metadata(
-    const grpcwrap_batch_context* ctx) {
-  return &(ctx->recv_initial_metadata);
+
+GPR_EXPORT void GPR_CALLTYPE
```
BusyJay, line 283: Can you separate all changes to C codes as a standalone PR? I think this part is good enough to be merged first. This is also a get around to the CI trust issue.

src/metadata.rs
```diff
@@ -254,6 +266,12 @@ impl Drop for Metadata {
     }
 }
 
+impl AsMut<grpc_metadata_array> for Metadata {
```
BusyJay, line 269: How about implementing it as `pub(crate) fn as_mut_ptr(&mut self) -> *mut grpc_metadata_array`?

src/task/promise.rs
```diff
@@ -58,7 +90,11 @@ impl Batch {
             if succeed {
                 let status = self.ctx.rpc_status();
                 if status.code() == RpcStatusCode::OK {
-                    guard.set_result(Ok(None))
+                    guard.set_result(Ok(BatchResult::new(
+                        self.ctx.recv_message(),
```
BusyJay, line 94: This should always be `None`.

BusyJay (commented):
Any tests to cover the new APIs?

src/task/promise.rs
christian-oudard: Thanks for catching this.

src/metadata.rs
christian-oudard: I agree, this is cleaner.

src/call/client.rs
christian-oudard: Not sure I understand your comment. This is used here:
```
    pub async fn headers(&mut self) -> Result<Metadata> {
        Ok(Pin::new(&mut self.imp.headers_f).await?.initial_metadata)
    }
```

src/call/mod.rs
```diff
@@ -286,6 +286,26 @@ impl BatchContext {
         let buf = self.take_recv_message()?;
         Some(GrpcByteBufferReader::new(buf))
     }
+
+    pub fn initial_metadata(&self) -> Metadata {
```
hunterlxt, line 290: add comments

src/call/mod.rs
```diff
@@ -286,6 +286,26 @@ impl BatchContext {
...
+        unsafe {
+            grpcio_sys::grpcwrap_batch_context_take_recv_initial_metadata(
+                self.ctx,
+                (&mut res).as_mut_ptr(),
+            );
+        }
+        res
+    }
+
+    pub fn trailing_metadata(&self) -> Metadata {
```
hunterlxt, line 301: same here

grpc-sys/grpc_wrap.cc
```diff
@@ -279,10 +279,19 @@ grpcwrap_request_call_context_destroy(grpcwrap_request_call_context* ctx) {
...
+
+  /* According to the documentation for struct grpc_op in grpc_types.h,
+   * ownership of keys and values for
+   * metadata stays with the call object. This means we have ref each of the
+   * keys and values here. */
+  size_t i;
+  for (i = 0; i < res->count; i++) {
+    grpc_slice_ref(res->metadata[i].key);
+    grpc_slice_ref(res->metadata[i].value);
+  }
```
hunterlxt, line 294: where to unref

hunterlxt (commented):
add comments on most pub functions

hunterlxt (commented):
seem that no integrated tests you can ref cases::metadata::test_metadata

BusyJay:
@christian-oudard do you have time to address comments? If not, I can help you to fix comments and get it merged.

christian-oudard:
> @christian-oudard do you have time to address comments? If not, I can help you to fix comments and get it merged.

Right now I don't have very much time to work on this. Last I looked at this PR, I was blocked on trying to find a place to split the C changes from the API changes. The differences in `grcp_wrap.cc` cause significant changes to internal function signatures, and I wasn't able to find an obvious place to stop propagating these changes.

BusyJay:
I continue your work in #555. I don't push directly to your branch as it may be used for your business.

BusyJay:
Close as #555 is merged.

PR #544, opened 2021-09-13
Update prost to 0.8

pingyu:
Update Prost to 0.8, to avoid security issue of 0.7 (RUSTSEC-2021-0073).
Related issue: tikv/tikv#10905, https://github.com/tikv/protobuf-build/pull/55

Note that `grpcio-proto` still depends on `prost 0.7` because of its build-dependency to `protobuf-build 0.12`. As `protobuf-build 0.12` depends on `grpcio-compiler 0.9`, there is a circular dependency between `grps-rs` and `protobuf-build`.

I think we should release `grps-rs` 0.10.0 first, then `protobuf-build` 0.13, and at last update `grpcio-proto` to depend on `protobuf-build` 0.13.

Signed-off-by: pingyu <shui.yu@126.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.10.0"
+version = "0.9.0"
 edition = "2018"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
@@ -53,4 +53,4 @@ use-bindgen = ["grpcio-sys/use-bindgen"]
 travis-ci = { repository = "tikv/grpc-rs" }
 
 [patch.crates-io]
-grpcio-compiler = { path = "compiler", version = "0.10.0", default-features = false }
+grpcio-compiler = { path = "compiler", version = "0.9.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.9.0"
+version = "0.10.0"
 edition = "2018"
 authors = ["The TiKV Project Developers"]
 license = "Apache-2.0"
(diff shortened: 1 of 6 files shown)


sticnarf:
Well, I think it is okay to release protobuf-build 0.13 first because it allows all recent `grpcio-compiler` versions.

Then, we can release v0.10 of `grpc-rs` and `grpcio-compiler` together. 

@BusyJay What do you think?

BusyJay:
> Well, I think it is okay to release protobuf-build 0.13 first because it allows all recent `grpcio-compiler` versions.
> 
> Then, we can release v0.10 of `grpc-rs` and `grpcio-compiler` together.

Agree. In long term, we better land #545 to solve the problem completely.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.9.0"
+version = "0.10.0"
```
BusyJay, line 3: Better keep the old version. We will send a standalone PR to mutate the version and update changelog.

BusyJay:
Is there any risk to upgrade to 0.8? I can see a comment here: https://github.com/tokio-rs/prost/issues/526#issuecomment-920886300.

pymongo:
> Is there any risk to upgrade to 0.8? I can see a comment here: [tokio-rs/prost#526 (comment)](https://github.com/tokio-rs/prost/issues/526#issuecomment-920886300).

prost 0.8 Timestamp need TryInto to convert to std::time::SystemTime, before 0.7 can use Into trait to convert

but I think this convert is safe on Linux in prost 0.8


prost <= 0.7 is compile failed on rustc >= 1.56.0

Cargo.toml
pingyu: Comment addressed. And change version of depending `protobuf-build` to `>=0.12`, to use the coming `0.13` version without modification again.
PTAL~

BTW, why better keep the old version ?

BusyJay: Because we may not just include one change in next minor version. This is also friendly for people who want to try the change using patch.

pingyu: Got it. Thanks~

BusyJay (approved):

hunterlxt (approved):

PR #547, opened 2021-09-18
tests: fix flaky client_send_all case

BusyJay:
Because the service returns early, it's possible the sink finish with
errors.


hunterlxt (approved):
LGTM

PR #550, opened 2021-10-01
*: support querying channelz

BusyJay:
This is a basic support for [channelz][1] that allows querying the
information using APIs. Full support is difficult because no library
supports parse JSON to protobuf in Rust.

[1]: https://github.com/grpc/proposal/blob/master/A14-channelz.md


BusyJay:
@NingLin-P This can be helpful for analyzing why there are so many connections.

BusyJay:
> ...Do implementations of other languages support returning content in proto format directly?

Yes, they have reflections and support converting between formats.

PR #554, opened 2021-12-29
Reduce dependency on futures crate

robjtede:
Reduces dependency graph of each crate:

| |before|after|
-|-|-
grpcio|50|30
benchmark|120|102
grpcio-compiler|2|2
grpcio-health|54|35
interop|71|51
grpcio-proto|54|33
tests-and-examples|63|57

A particularly important change has been made to the `compiler` crate which now references the `Future` trait directly from `std` rather than it's re-export through the futures crate, meaning that downstream crates can also move away from having a direct dependency on `futures`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,8 @@ all-features = true
 [dependencies]
 grpcio-sys = { path = "grpc-sys", version = "0.9", default-features = false }
 libc = "0.2"
-futures = "0.3"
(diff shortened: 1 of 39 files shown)


BusyJay (approved):
Thanks!

BusyJay (approved):

hunterlxt (approved):

PR #555, opened 2022-01-21
Server-side metadata for all call types

BusyJay:
This PR continues #543.


christian-oudard:
Thank you so much for continuing this work, I've been busy with other priorities.

hunterlxt (approved):
Looks almost the same as the original #543 except c++ wrapper.

hunterlxt (dismissed):
LGTM

PR #558, opened 2022-02-25
Update grpc c core to 1.44.0

### LINKED ISSUE #557: Openssl feature pulls in boringssl
**Describe the bug**
Enabling the openssl feature pulls in 10MB of boringssl-src.
This is due to:
https://github.com/tikv/grpc-rs/blob/472ae8ac6dee4702b46d65ecac35cda0f333a64a/grpc-sys/Cargo.toml#L57-L58

**Expected behavior**
IIUC BoringSSL is not required since we're using openssl instead, so that should not be added as a dependency.

BusyJay:
This PR also uses the same bindings for both (x86_64|aarch64)_(macos|linux). CI will check if it's OK to do so. This should make maintenance easier.

Changes from #539 is also included in this PR.

Features `use-bindgen` is removed as it can be detected during compile time now.

This PR also rename the secure features to boringssl for better understanding. And now only enabling openssl features will not download boringssl anymore. Close #557.


grpc-sys/build.rs
```diff
@@ -19,12 +19,16 @@ const GRPC_VERSION: &str = "1.38.0";
 #[rustfmt::skip]
 const COMMON_DEPS: &[&str] = &[
     "absl_bad_optional_access", "absl_bad_variant_access", "absl_base", "absl_city", "absl_civil_time",
-    "absl_cord", "absl_debugging_internal", "absl_demangle_internal", "absl_exponential_biased",
-    "absl_graphcycles_internal", "absl_hash", "absl_hashtablez_sampler", "absl_int128", "absl_log_severity",
-    "absl_malloc_internal", "absl_raw_hash_set", "absl_raw_logging_internal", "absl_spinlock_wait",
-    "absl_stacktrace", "absl_status", "absl_statusor", "absl_str_format_internal", "absl_strings",
+    "absl_cord", "absl_cord_internal", "absl_cordz_functions", "absl_cordz_handle", "absl_cordz_info",
```
hunterlxt, line 22: can you post the command to run pkg-config manually into the comment above

grpc-sys/build.rs
```diff
@@ -379,38 +391,47 @@ fn bindgen_grpc(file_path: &Path) {
...
+// need to be updated by default unless the _gen-bindings feature is specified.
 // Other platforms use bindgen to generate the bindings every time.
 fn config_binding_path() {
     let target = env::var("TARGET").unwrap();
     let file_path: PathBuf = match target.as_str() {
-        "x86_64-unknown-linux-gnu" | "aarch64-unknown-linux-gnu" => {
+        "x86_64-unknown-linux-gnu"
+        | "aarch64-unknown-linux-gnu"
+        | "x86_64-apple-darwin"
+        | "aarch64-apple-darwin" => {
```
hunterlxt, line 402: Why are the bindings the same across this platforms?

BusyJay: I guess because they are both unix-like system and share many common designs. The windows bindings is different from linux though. For example, windows use i32 for enum while unix use u32. More to see https://github.com/rust-lang/rust-bindgen/issues/1907

hunterlxt: I mean, did you found it unconsciously, or did you get the result via bindgen config.

hunterlxt (dismissed):
Most LGTM

grpc-sys/build.rs
BusyJay: I found it when I'm working on https://github.com/BusyJay/tirocks/pull/1.

hunterlxt (approved):

PR #566, opened 2022-03-26
metadata: avoid change metadata ref

BusyJay:
After metadata is received, every entry should be owned by the grpc c
core, application should clone the byte slice if necessary.

In the past, we get around the problem by increasing the reference
count of the slice. However, it's unsafe to do so as the slice is not
guaranteed to be accessible in the first place.

This PR fixes the problem by introducing an unowned metadata type. It
works just like Metadata, but accessing its content requires manual
check for the lifetime of associated call.

I tried to add test case to cover the unsafe access, however it's too racy
and nearly impossible to introduce a case without touching the C core.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -147,7 +147,7 @@ impl MetadataBuilder {
 ///
 /// Metadata value can be ascii string or bytes. They are distinguish by the
 /// key suffix, key of bytes value should have suffix '-bin'.
-#[repr(C)]
+#[repr(transparent)]
 pub struct Metadata(grpc_metadata_array);
 
 impl Metadata {
@@ -273,7 +273,7 @@ unsafe impl Sync for Metadata {}
 ///
 /// gRPC C Core manages metadata internally, it's unsafe to read them unless
 /// call is not destroyed.
-#[repr(C)]
+#[repr(transparent)]
 pub struct UnownedMetadata(grpc_metadata_array);
 
 impl UnownedMetadata {
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -234,10 +234,6 @@ impl Metadata {
         }
         &[]
     }
-
-    pub(crate) fn as_mut_ptr(&mut self) -> *mut grpc_metadata_array {
-        &mut self.0 as _
-    }
 }
 
 impl fmt::Debug for Metadata {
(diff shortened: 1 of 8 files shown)


BusyJay:
Long time tests show that https://github.com/tikv/tikv/issues/12202 and https://github.com/tikv/tikv/issues/12198 are resolved by this patch.

hunterlxt:
can you explain `However, it's unsafe to do so as the slice is not
guaranteed to be accessible in the first place.`?

src/metadata.rs
```diff
@@ -273,6 +269,38 @@ impl Drop for Metadata {
...
+#[repr(C)]
+pub struct UnownedMetadata(grpc_metadata_array);
+
+impl UnownedMetadata {
+    #[inline]
+    pub fn empty() -> UnownedMetadata {
+        unsafe { mem::transmute(Metadata::with_capacity(0)) }
+    }
+    #[inline]
+    pub unsafe fn assume_valid(&self) -> &Metadata {
```
hunterlxt, line 285: maybe `convert_to_owned` better

src/metadata.rs
```diff
@@ -273,6 +269,38 @@ impl Drop for Metadata {
...
+
+    pub fn as_mut_ptr(&mut self) -> *mut grpc_metadata_array {
+        &mut self.0 as _
+    }
+}
+
+impl Drop for UnownedMetadata {
+    #[inline]
+    fn drop(&mut self) {
+        unsafe { grpcio_sys::grpcwrap_metadata_array_destroy_metadata_only(&mut self.0) }
```
hunterlxt, line 297: why free here because I don't see any pointer init

src/metadata.rs
BusyJay: It's not a convertion. It's more like a as_ref/deref. The returned value is not owned, it just reuses the struct for accessing APIs. That's why a reference is returned.

src/metadata.rs
BusyJay: It's initialized at https://github.com/tikv/grpc-rs/pull/566/files#diff-1d33dfbea0d58c789cffd727c3a0b6e1fff524df008f91541682196e9c3c5665R298 and https://github.com/tikv/grpc-rs/pull/566/files#diff-1d33dfbea0d58c789cffd727c3a0b6e1fff524df008f91541682196e9c3c5665R313.

BusyJay:
> can you explain that?

The order of handling batch result and destroying a call is uncertain. So it's possible that when the batch result is returned from the queue_next, the corresponding grpc call is already destroyed. In that case, referencing internal slices is UAF, and destroy the metadata later will dereference it, which is a double free.

src/metadata.rs
```diff
@@ -147,7 +147,7 @@ impl MetadataBuilder {
 ///
 /// Metadata value can be ascii string or bytes. They are distinguish by the
 /// key suffix, key of bytes value should have suffix '-bin'.
-#[repr(C)]
+#[repr(transparent)]
```
hunterlxt, line 150: seem that all struct with single field using `repr(C)` can be replaced with `repr(transparent)`.

hunterlxt (approved):
LGTM

PR #573, opened 2022-05-24
feat: Add support for GRPC_ARG_ENABLE_HTTP_PROXY parameter

keskydi:
Add the support for the GRPC_ARG_ENABLE_HTTP_PROXY parameter in order to communicate between local processes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.10.2 - 2022-04-15
+
+- Make `ResourceQuota` cloneable (#568)
+- Allow use local subchannel pool (#565)
+
 # 0.10.1 - 2022-03-28
 
 - Fix potential UAF and double free (#566)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpcio"
-version = "0.10.1"
+version = "0.10.2"
 edition = "2018"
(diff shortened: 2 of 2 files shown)


src/channel.rs
```diff
@@ -295,6 +295,15 @@ impl ChannelBuilder {
         );
         self
     }
+    
+    /// If set to zero, disables use of http proxies.
+    pub fn enable_http_proxy(mut self, num: i32) -> ChannelBuilder {
```
BusyJay, line 300: How about using `bool`?

BusyJay (commented):
Thanks! You need to sign off your commit.

src/channel.rs
keskydi: The Options enum only define Integer, String and Pointeur

keskydi: modification done

BusyJay (approved):

BusyJay:
You can use command `git commit --signoff` to signoff your commits. Note all commits need to be signed off.

keskydi:
I had to fight with git to sign all my commits so i recreated a new pull request : https://github.com/tikv/grpc-rs/pull/574

keskydi:
Add trouble with signing so i ended up doing another pull request : https://github.com/tikv/grpc-rs/pull/575

PR #577, opened 2022-06-09
Add new Gzip param

zkkxu:
add initial arguments for gzip compression:
            gzip_compression_level: represent gzip compression level, the origin gzip compression level is 6 and hardcoding;
            compression_lower_bound: this represent gzip will compress the data only larger than this

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,6 @@ jobs:
           - host: ubuntu-latest
             profile: 
             suffix:
-          - host: ARM64
-            profile: --release
-            suffix: -Arm64
     name: Linux-Stable${{ matrix.suffix }}
     runs-on: ${{ matrix.host }}
     steps:
@@ -62,9 +59,6 @@ jobs:
           - host: ubuntu-latest
             profile: 
             suffix:
-          - host: ARM64
-            profile: --release
-            suffix: -Arm64
     name: Linux-Stable-openssl${{ matrix.suffix }}
     runs-on: ${{ matrix.host }}
     steps:


src/channel.rs
```diff
@@ -314,6 +314,24 @@ impl ChannelBuilder {
         self
     }
 
+    /// Set default gzip compression level.
+    pub fn default_gzip_compression_level(mut self, level: usize) -> ChannelBuilder {
```
BusyJay, line 318: We should make this a nightly API as it's not supported by upstream.

src/channel.rs
```diff
@@ -314,6 +314,26 @@ impl ChannelBuilder {
...
+        self.options.insert(
+            Cow::Borrowed(grpcio_sys::GRPC_GZIP_COMPRESSION_LEVEL),
+            Options::Integer(level as i32),
+        );
+        self
+    }
+
+    /// Set default grpc min message size to compression.
+    #[cfg(feature = "nightly")]
+    pub fn default_grpc_min_message_size_to_compress(mut self, lowerBound: usize) -> ChannelBuilder {
```
BusyJay, line 329: Any test case?

zkkxu: had to mock, and is just a options

src/channel.rs
zkkxu: done

src/channel.rs
```diff
@@ -733,3 +753,23 @@ impl Channel {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+    use crate::env::Environment;
+    use crate::ChannelBuilder;
+
+    #[test]
+    fn test_grpc_min_message_size_to_compress() {
```
BusyJay, line 764: I think they need to be guarded by `nightly` feature.

BusyJay (commented):
And .github/workflows/ci.yml needs to be updated.

src/channel.rs
zkkxu: done

BusyJay (approved):

PR #580, opened 2022-06-28
*: remove arm ci check

BusyJay:
Confirmed from pingcap that arm self hosted runner is not available anymore, so we need to remove these checks in CI as they will always timeout.


zhangjinpeng1987 (approved):
LGTM

PR #586, opened 2022-09-10
*: bump 0.11.0

BusyJay:


BusyJay:
Because this upgrade doesn't involve any code change, perhaps I can publish it as minor version change instead next time. 🤔

PR #588, opened 2022-09-14
Update README.md

dertin:
add git submodule for third party deps

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -19,6 +19,7 @@ $ git clone https://github.com/pingcap/grpc.git
 
 ```
 $ cd grpc-rs
+$ git submodule update --init --recursive
 $ cargo build -p benchmark --release
 ```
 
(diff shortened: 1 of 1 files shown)


benchmark/README.md
```diff
@@ -19,13 +19,15 @@ $ git clone https://github.com/pingcap/grpc.git
 
 ```
 $ cd grpc-rs
+$ git submodule update --init --recursive
```
BusyJay, line 22: We have customize step `cargo xtask submodule`. It will only clone needed submodules.

dertin: I have had an error due to missing dependencies and I had to use git submodule to obtain them

BusyJay: Do you mean `cargo xtask submodule` fail? Can you paste the error here?

dertin: I didn't see in the README that I should use `cargo xtask submodule`

That's why I suggested to use `git submodule update --init --recursive` 
should include  `cargo xtask submodule` in the README

BusyJay: Yes, I agree there should be instructions to initialize the submodules in README.md. I just suggest to use `cargo xtask submodule` instead of `it submodule update --init --recursive`.

> should include cargo xtask submodule in the README

That's exact what I mean.

BusyJay:
Solved by #594

PR #593, opened 2022-10-24
Update grpc 1.46.5

BusyJay:
1.46.x is the last patch that support C++-11. So update to this version first.

### DIFF
--- a/src/security/credentials.rs
+++ b/src/security/credentials.rs
@@ -185,15 +185,14 @@ impl ServerCredentialsBuilder {
 
     /// Finalize the [`ServerCredentialsBuilder`] and build the [`ServerCredentials`].
     pub fn build(self) -> ServerCredentials {
-        let credentials = unsafe {
+        unsafe {
             let opt = grpcio_sys::grpc_ssl_server_credentials_create_options_using_config(
                 self.cer_request_type.to_native(),
                 self.build_config(),
             );
-            grpcio_sys::grpc_ssl_server_credentials_create_with_options(opt)
-        };
-
-        ServerCredentials { creds: credentials }
+            let credentials = grpcio_sys::grpc_ssl_server_credentials_create_with_options(opt);
+            ServerCredentials::from_raw(credentials)
+        }
     }
 }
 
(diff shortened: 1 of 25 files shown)


tabokie:
The benchmark results are a little confusing. I assume "Diff (engine)" is the latency comparison between two versions with the specific engine type. Then why can we still test epollex when it is removed from 1.46.5?

BusyJay:
The diff is throughput difference. Diff (epollex) compares 1.46.5(epoll1) with 1.44.0(epollex), diff (epoll1) compares 1.46.5(epoll1) with 1.44.0(epoll1). Testing both two results to show that major regression is from engine change.

Giving that there is a high risk of regression, but my benchmark of current TiKV using 1.44.0 and epoll1 has no obvious regression, and even some improvement under certain benchmark type(like bank). That is also part of reason why I want to upgrade grpc and check for more serious benchmark types.

src/security/credentials.rs
```diff
@@ -209,29 +208,28 @@ impl Drop for ServerCredentialsBuilder {
...
-            grpc_sys::grpc_server_credentials_release(self.creds);
+            let opt = grpcio_sys::grpc_ssl_server_credentials_create_options_using_config_fetcher(
+                cer_request_type.to_native(),
+                Some(server_cert_fetcher_wrapper),
+                fetcher_wrap_ptr as _,
+            );
+            let mut creds = ServerCredentials::from_raw(
+                grpcio_sys::grpc_ssl_server_credentials_create_with_options(opt),
+            );
+            creds._fetcher = Some(Box::from_raw(fetcher_wrap_ptr));
```
tabokie, line 231: `_fetcher` isn't fully gated with `_secure`?

BusyJay: No, grpc c core will not delete it.

BusyJay: Oh, you mean whether it can be accessed in "credentials.rs". Yes, it can. The file is also feature gated by "_secure".

tabokie (approved):

PR #595, opened 2022-10-31
*: update prost to 0.11

### LINKED ISSUE #545: remove protobuf-build completely
grpcio-proto depends on protobuf-build to build dependency, protobuf-build depends on grpcio-compiler, hence cause circular dependency. It's not easy to maintain.

Remove protobuf-build completely. Instead, using custom scripts to generate files instead.

BusyJay:
Since 0.11, prost-build never bundles protoc binary anymore. To avoid requiring protoc for grpcio-proto users, we bundle the sources instead of generating at compile time. This can lead to potential problems like:
- Miss match version between generating tools and runtime
- Repository becomes bloated

But there seems to be no better alternative.

Since this PR when using prost with grpcio, protoc is also required in PATH.

This PR also removes protobuf-build dependencies. From now on, it will be easier to publish new version of grpcio-compiler.

Close #545.


health/src/proto/protobuf/health.rs
```diff
@@ -443,3 +443,5 @@ pub fn file_descriptor_proto() -> &'static ::protobuf::descriptor::FileDescripto
         })
     }
 }
+
+pub use super::health_grpc::*;
```
sticnarf, line 447: I'm curious about the small structure change. Is this line added by hand?

BusyJay: No, it's added by https://github.com/tikv/grpc-rs/pull/595/files#diff-9c5e61698d6d8a8ef4a3da7075c31857769134a932ac141d4eb62f68f131aa35R176-R188

sticnarf (approved):

PR #597, opened 2022-11-08
client: allow accessing channel

BusyJay:
We have supported channel state, but it's very inconvenient without the capability of accessing underlying channel of a client.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -98,4 +98,10 @@ impl Client {
         let kicker = self.kicker.clone();
         Executor::new(self.channel.cq()).spawn(f, kicker)
     }
+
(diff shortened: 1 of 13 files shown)


tabokie (approved):

src/client.rs
```diff
@@ -98,4 +98,10 @@ impl Client {
         let kicker = self.kicker.clone();
         Executor::new(self.channel.cq()).spawn(f, kicker)
     }
+
+    /// Get the underlying channel.
+    #[inline]
+    pub fn channel(&self) -> &Channel {
```
tabokie, line 104: Needs to update codegen.

PR #603, opened 2022-12-30
Provide more debugging info for RpcStatus

rshearman:
* grpc-sys: provide error string for receive status on client calls
* call: provide debug error string in RpcStatus
* ~error: make Display ouput more human-readable~
* ~call: make Display of RpcStatusCode more human-readable~

Example of contents of `RpcStatus::debug_error_string`:
```
thread 'test_health_watch_multiple' panicked at 'called `Result::unwrap()` on an `Err` value: RpcFailure(RpcStatus { code: 14-UNAVAILABLE, message: "failed to connect to all addresses", details: [], debug_error_string: "{\"created\":\"@1672417943.726323715\",\"description\":\"Failed to pick subchannel\",\"file\":\"/home/rob/grpc-rs/grpc-sys/grpc/src/core/ext/filters/client_channel/client_channel.cc\",\"file_line\":3217,\"referenced_errors\":[{\"created\":\"@1672417943.726320053\",\"description\":\"failed to connect to all addresses\",\"file\":\"/home/rob/grpc-rs/grpc-sys/grpc/src/core/lib/transport/error_utils.cc\",\"file_line\":165,\"grpc_status\":14}]}" })', health/tests/health_check.rs:137:5
```

### DIFF
--- a/src/call/mod.rs
+++ b/src/call/mod.rs
@@ -3,6 +3,7 @@
 pub mod client;
 pub mod server;
 
+use std::ffi::CStr;
 use std::fmt::{self, Debug, Display};
 use std::future::Future;
 use std::pin::Pin;
@@ -166,6 +167,9 @@ pub struct RpcStatus {
     ///
     /// See also https://grpc.io/docs/guides/error/#richer-error-model.
     details: Vec<u8>,
+
+    /// Debug error string
+    debug_error_string: String,
 }
 
 impl Display for RpcStatus {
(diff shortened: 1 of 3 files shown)


src/call/mod.rs
```diff
@@ -71,28 +66,41 @@ macro_rules! status_codes {
                 )
             }
         }
+
+        impl Display for RpcStatusCode {
```
BusyJay, line 70: What's the difference?

BusyJay (commented):
The debug_string part LGTM, the rest seems unnecessary.

src/call/mod.rs
rshearman: The numeric value of the error code isn't included  and a human-readable version of the status code symbol is given (i.e. lowercase and underscores replaced with spaces). E.g. Instead of giving "15-DATA_LOSS", it gives "data loss". The former is great for the Debug trait implementation, but IMHO could be improved when being presented to a user, which is the intention of the Display implementation.

rshearman:
> The debug_string part LGTM, the rest seems unnecessary.

Could I ask why? As an example, currently, we are displaying an error to a user and the Display trait for Error outputs it as:

`RpcFailure: 7-PERMISSION_DENIED Unknown or deactivated device`

IMHO it would be nicer and easier to understand for the user to output it as:

`RPC failure: Unknown or deactivated device`

Which seems to better match the intent of the Display trait being intended for "user-facing output" (https://doc.rust-lang.org/std/fmt/trait.Display.html).

rshearman:
>  I suppose the user here means developers. 

That's not how I interpret the use of the Display trait versus the Debug trait, i.e. as a developer I would look at Debug rather than Display.

> If you want to display a "better" message to normal users like users that use your application, then you may consider change the Display implementation of error type defined in your application instead of leverage the implementation to the library directly. As "better" is a relative concept in different context. One format may seem better in your case while not in others.

I can absolutely do that, but my motivation here was to give others the benefit of the change. I understand there is a tradeoff, but I was hoping in making a bigger distinction between the Debug and Display trait implementations that applications can choose one or the other depending on their needs.

If I haven't been able to convince you that the Display implementation changes are worthwhile, I'll remove them from this PR - just let me know.

BusyJay:
The target users of the library are developers, so we better keep the original format. :)

rshearman:
I've removed the changes to the Display trait implementations.

rshearman:
Various CI jobs are failing, but these appear to match a scheduled run without my changes, e.g.: https://github.com/tikv/grpc-rs/actions/runs/3803392868

BusyJay (approved):

PR #608, opened 2023-02-02
*: fix clippy warnings

BusyJay:


sticnarf:
`uninlined_format_args` is downgraded to pedantic according to https://blog.rust-lang.org/2023/02/09/Rust-1.67.1.html. Should we enforce them now?

BusyJay:
It's downgraded temporary. If it's going to be upgraded again, the work will need to be do again, so better land this change instead.

sticnarf (approved):
LGTM

PR #611, opened 2023-02-17
Remove unnecessary warnings_into_errors

puzzle-rusher:
I ran into a problem when I tried to compile this project. `cc` stops compilation when 184 warnings are encountered, while allowing a maximum of 20 errors.
`Warning: turning warnings into errors only make sense if you are a developer of the crate using cc-rs.
Some warnings only appear on some architecture or specific version of the compiler. Any user of this crate, or any other crate depending on it, could fail during compile time.`

### DIFF
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -526,7 +526,6 @@ fn main() {
         cc.flag("-std=c++11");
     }
     cc.file("grpc_wrap.cc");
-    cc.warnings_into_errors(true);
     cc.compile("libgrpc_wrap.a");
 
     config_binding_path();


BusyJay:
grpc_wrap.cc is maintained by the project, it's expected to be no warnings. Can you paste the actual warnings?

puzzle-rusher:
> grpc_wrap.cc is maintained by the project, it's expected to be no warnings. Can you paste the actual warnings?

https://gist.github.com/puzzle-rusher/fee086d8b370965c76134c7943bd745b

puzzle-rusher:
> grpc_wrap.cc is maintained by the project, it's expected to be no warnings. Can you paste the actual warnings?

I don't think this is wrong behavior anyway, if the `cc` library warns you of this possibility, you should heed them. I spent a lot of time to find the reason why I can't compile my project which has a transitive dependency on your project, I don't want anyone to waste time on this.

BusyJay:
The warnings are generated by the builtin headers instead of any code from grpc-rs. It means your compiler might not match the headers. I can see there is a similar issue in other project, maybe you can try similar solutions. https://github.com/golang/go/issues/38876

puzzle-rusher:
> The warnings are generated by the builtin headers instead of any code from grpc-rs. It means your compiler might not match the headers. I can see there is a similar issue in other project, maybe you can try similar solutions. [golang/go#38876](https://github.com/golang/go/issues/38876)

It's a common situation. And it's not the reason to prohibit using your library, don't you think?

Issue #584, opened 2022-08-10
Upgrade protobuf 3


dertin:
Protobuf 3 is current stable version. Compared to version 2 it implements:

- runtime reflection
- JSON and text format parsing and printing
- dynamic messages (messages which can be created from .proto file on the fly without code generation)

https://github.com/stepancheg/rust-protobuf/#versions-and-branches

Is this update on the roadmap?

BusyJay:
No. I'm happy to accept a PR if anyone want to contribute. Just make sure people can still using protobuf 2 via features.

djc:
@BusyJay curious, why not?

BusyJay:
We were lack of resource at that time. I will work on this these days if no one wants to give a try.

djc:
Do you still need to keep protobuf 2 around via features, or will that be changed eventually, too?

BusyJay:
Yes, I will keep pb 2 via features. Though may enable pb 3 by default.

BusyJay:
I can see rust-protobuf is not active for the past months. And the README also calls for help for maintenance.

The reason why we stick to rust-protobuf 2 is that it has been on production for quite a long time, it seems a more stable version unless we require the new features.

Also, protobuf team seems to be implementing their own Rust version: https://github.com/protocolbuffers/protobuf/commit/89d6b74dea804995488980c44d3f53855e926d69.

I would like to receive more feedbacks on this issue.

djc:
I mostly use tonic/prost at work -- mainly here as a maintainer for opentelemetry-rust which in a number of crates supports both gRPC stacks. Curious if you think it will ever make sense for TiKV to migrate to tonic/prost?

BusyJay:
No, TiKV may probably stick to grpc-rs. Switching to other protobuf implementation is more likely compared to switching grpc implementation.

andrewbaxter:
Looks like this may have been resolved in https://github.com/tikv/grpc-rs/pull/615 ?

BusyJay:
Yes.

PR #615, opened 2023-04-05
Protobuf 3 support

lb034582341:
Hello, first time contributor here. Following [#584](https://github.com/tikv/grpc-rs/issues/584#issuecomment-1210697454), here's a first draft of the migration to protobuf 3.2. It's a work-in-progress and we'll surely need to iterate to find a good solution.

I haven't gated the changes behind a feature yet (as requested [here](https://github.com/tikv/grpc-rs/issues/584#issuecomment-1427697543)). I'm not exactly sure how to do this yet.

Also, it seems that for protobuf 3, the functionality of the `compiler/` folder has been factored out in the [protobuf-codegen crate](https://crates.io/crates/protobuf-codegen). Do we want to keep the folder and call the crate, or update the documentation to directly use the crate ?
It's been temporarily removed because it doesn't compile with protobuf 3.2.

The diff is large but only [Protobuf 3.2: syntax changes](https://github.com/tikv/grpc-rs/pull/615/commits/c535cf426fb13c588ed9d11d985637530eba44cb) contains meaningful changes (and most of them are trivial).

I have ran tests + clippy successfully.

Feedback is welcome, thanks !


lb034582341:
Hello, just checking in to see if there's interest in pushing this :) thanks

BusyJay:
Thanks for your contributions! Sorry for the delay, I will take a look in following 3 days.

lb034582341:
No rush at all on my side! Knowing that it might move forward at a later point is good for me. thanks

BusyJay:
> I haven't gated the changes behind a feature yet

From what I can see from this PR, there are three type of changes:
- API change
    We can introduce a compatible layer in grpc-rs so that no matter 2.x or 3.x is used, we can still using the same API sets. For example, `v.field = Some(v).into()` should work for both 2.x and 3.x, we don't need to change that; `CodedInputStream::from_buffered_reader` and `CodedInputStream::from_buf_read` don't work with each other, we need to introduce a function to invoke diffferent methods when different feature is enabled.
- Generated code change
    Unfortunately, we can do little thing about it. We may need to generate files for both versions and check in to the version control.
- Compiler change
    We may also introduce a compatible layer and only expose necessary common APIs. The tricky part here is TiKV project has customized the compiler to provide functionalities like log redact. So we may need to keep using protobuf for v2 and use protobuf-codegen only for v3.

lb034582341:
Hi BusyJay, sorry for the long wait. Thanks for the suggestions, they have been implemented.

The upgrade is gated behind a new feature `protobufv3-codec` which behaves like `protobuf-codec`. It is mutually exclusive with it, and also `prost`. Just a heads-up, you are probably aware that this [might not be recommended](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features), but I followed the logic of the existing code.

No changes have been made to the defaults. To build/test with protobuf v3:
`cargo test --features protobufv3-codec --no-default-features`

As discussed a lot of changes come from the generated files for v3, which can be ignored. [This commit](https://github.com/tikv/grpc-rs/pull/615/commits/75016e7c27df311bc6e7fcf8fdad8d491a9e4563) regroups the changes to the various `Cargo.toml`, and [this commit](https://github.com/tikv/grpc-rs/pull/615/commits/8609072ad5b5c99a6f8f6737b69d42629c8d1dff) contains the actual rust code changes.

`cargo test` and `cargo clippy` have been ran successfully.

thanks for your time!

health/src/service.rs
```diff
@@ -14,13 +14,39 @@ use std::task::{Context, Poll, Waker};
...
+    use crate::proto::ServingStatus;
+    pub const NOT_SERVING: ServingStatus = ServingStatus::NotServing;
+    pub const SERVICE_UNKNOWN: ServingStatus = ServingStatus::ServiceUnknown;
+}
+
+#[cfg(feature = "protobufv3-codec")]
+mod constants {
+    use crate::proto::health_check_response::ServingStatus;
+    pub const NOT_SERVING: ServingStatus = ServingStatus::NOT_SERVING;
+    pub const SERVICE_UNKNOWN: ServingStatus = ServingStatus::SERVICE_UNKNOWN;
```
mgeisler, line 34: I think you might be able to get away with just

```suggestion
    pub use crate::proto::health_check_response::ServingStatus::*;
```

mgeisler: Similar for other places where you use this trick. Alternatively, you could remove the `constants` module completely and do something like this in the branch above — I actually think that's even simpler:

```rust
    use crate::proto::ServingStatus::NotServing as NOT_SERVING;
    use crate::proto::ServingStatus::ServiceUnknown as SERVICE_UNKNOWN;
```

interop/src/client.rs
```diff
@@ -290,3 +291,265 @@ impl Client {
         Ok(())
     }
 }
+
+
+#[cfg(feature = "protobufv3-codec")]
```
mgeisler, line 296: It looks like you could move this conditional down to the level of the methods — `new` looks like it's identical in the two branches.

src/codec.rs
```diff
@@ -57,13 +61,23 @@ pub mod pb_codec {
...
     }
 }
 
+#[cfg(feature = "protobuf-codec")]
+fn from_buf_read(reader: & mut MessageReader) -> protobuf::CodedInputStream {
+    return protobuf::CodedInputStream::from_buffered_reader(reader);
+}
+
+#[cfg(feature = "protobufv3-codec")]
+fn from_buf_read(reader: & mut MessageReader) -> protobufv3::CodedInputStream {
```
mgeisler, line 77: Since the function is a no-op, perhaps you could just do

```rust
use protobufv3::CodedInputStream::from_buf_read;
```

here to define the `from_buf_read` symbol with the right type?

compiler/src/codegen.rs
```diff
@@ -564,10 +597,16 @@ impl<'a> ServiceGen<'a> {
...
     }
 
+    #[cfg(feature = "protobuf-codec")]
     fn service_name(&self) -> String {
         util::to_camel_case(self.proto.get_name())
     }
 
+    #[cfg(feature = "protobufv3-codec")]
+    fn service_name(&self) -> String {
+        util::to_camel_case(self.proto.name)
```
mgeisler, line 607: You could move the conditional into the method here and avoid duplicating the method signature.

health/src/service.rs
lb034582341: Done, thanks !

interop/src/client.rs
lb034582341: I thought it would be cleaner by doing it at the Impl level, but you're right, this duplicates some code and adds stuff to maintain. Done, thanks

compiler/src/codegen.rs
```diff
@@ -678,6 +718,35 @@ fn gen_file(
     })
 }
 
+#[cfg(feature = "protobufv3-codec")]
+fn gen_file(
+    file: &FileDescriptorProto,
+    root_scope: &RootScope,
+) -> Option<compiler_plugin::GenResult> {
+    if file.service.is_empty() {
```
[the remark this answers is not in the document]
lb034582341: No, there's another reference to file.service in the for loop below.

compiler/src/codegen.rs
lb034582341: Done, thanks

src/codec.rs
lb034582341: It doesn't seem to work, `CodedInputStream` is a struct not a module.

lb034582341:
(Force-push for signoff, sorry)

Thanks a lot @mgeisler for the review!

compiler/src/codegen.rs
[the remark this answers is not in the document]
mgeisler: Okay, I would still suggest factoring this one variable out so that you don't duplicate the rest.

lb034582341:
Hello, I apologize for the ping... It's been a month, I'm just wondering what to do with my proposed changes ?  Thank you

lb034582341:
Sorry to insist @BusyJay, is there any way forward for this ? Someone else I could bother?
This is the last crate holding the whole Android codebase on protobuf 2. 
Thank you

benchmark/src/bench.rs
```diff
@@ -18,13 +18,22 @@ use grpc_proto::testing::services_grpc::BenchmarkService;
...
 
+#[cfg(feature = "protobuf-codec")]
 fn gen_resp(req: &SimpleRequest) -> SimpleResponse {
     let payload = util::new_payload(req.get_response_size() as usize);
     let mut resp = SimpleResponse::default();
     resp.set_payload(payload);
     resp
 }
 
+#[cfg(feature = "protobufv3-codec")]
```
BusyJay, line 29: Seems protobuf v2 can also use this logic as all fields in v2 are public. And there are also other places in this PR can use the same code in both v2 and v3.

lb034582341: Good catch, this makes the code simpler. It doesn't always work due to type differences, but I went over the changes and refactored what I could. Thank you for the review!

proto/src/proto.rs
```diff
@@ -217,14 +217,44 @@ pub mod protobuf {
...
+      pub mod payloads;
+      pub mod services;
+      pub mod services_grpc;
+      pub mod stats;
+      pub mod test;
+      pub mod test_grpc;
+  }
+}
+
+#[cfg(feature = "protobufv3-codec")]
```
BusyJay, line 232: How about:
```
#[cfg(any(feature = "A", feature = "B"))]
#[cfg_attr(feature = "A", path = "A")]
#[cfg_attr(feature = "B", path = "B")]
pub mod A {
    ...
}
```
?

interop/src/server.rs
```diff
@@ -132,28 +151,37 @@ impl TestService for InteropTestService {
...
                 }
                 // A workaround for timeout_on_sleeping_server test.
                 // The request only has 27182 bytes of zeros in payload.
                 //
                 // Client timeout 1ms is too short for grpcio. The server
                 // can response in 1ms. To make the test stable, the server
                 // sleeps 1s explicitly.
+
+                #[cfg(feature = "protobuf-codec")]
                 if req.get_payload().get_body().len() == 27182
```
BusyJay, line 175: You can move this condition to last so that don't need to repeat all other conditions.

tests-and-examples/tests/tests.rs
```diff
@@ -1,3 +1,7 @@
 // Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
 
+#[cfg(feature = "protobuf-codec")]
 mod cases;
+
+#[cfg(feature = "protobufv3-codec")]
+mod cases_protobufv3;
```
BusyJay, line 7: Tests should be updated in place. Otherwise it's hard to maintain two copies for every test cases.

interop/src/server.rs
lb034582341: Is this what you have in mind ?

proto/src/proto.rs
lb034582341: Done, thanks !

src/codec.rs
```diff
@@ -57,13 +61,23 @@ pub mod pb_codec {
...
+        let mut s = from_buf_read(&mut reader);
         let mut m = T::new();
         m.merge_from(&mut s)?;
         Ok(m)
     }
 }
 
+#[cfg(feature = "protobuf-codec")]
+fn from_buf_read(reader: & mut MessageReader) -> protobuf::CodedInputStream {
+    return protobuf::CodedInputStream::from_buffered_reader(reader);
```
BusyJay, line 73: No need to return in Rust.

tests-and-examples/examples_protobufv3/route_guide/util.rs
```diff
@@ -0,0 +1,83 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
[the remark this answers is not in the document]
lb034582341: Done

src/codec.rs
lb034582341: Fixed overall through the PR (except in auto-generated files in case someone re-generates them). thanks.

src/error.rs
```diff
@@ -84,18 +87,18 @@ pub type Result<T> = result::Result<T, Error>;
 
 #[cfg(all(test, feature = "protobuf-codec"))]
 mod tests {
-    use std::error::Error as StdError;
+  use std::error::Error as StdError;
```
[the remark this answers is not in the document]
lb034582341: Right. There were a couple of files that were not properly formatted, I ran `cargo fmt` on the repo. please let me know if you only want me to manually format the bits that I changed. Thanks

tests-and-examples/examples/route_guide/client.rs
```diff
@@ -82,18 +88,25 @@ async fn record_route(client: &RouteGuideClient) -> Result<()> {
     let (mut sink, receiver) = client.record_route()?;
     for _ in 0..10usize {
         let f = features.choose(&mut rng).unwrap();
+        #[cfg(feature = "protobuf-codec")]
         let point = f.get_location();
```
BusyJay, line 92: Why not `f.get_location().to_owned()`? So `sink.send` can use the same code.

tests-and-examples/Cargo.toml
```diff
@@ -6,8 +6,9 @@ autoexamples = false
 publish = false
 
 [features]
-default = ["protobuf-codec"]
+default = ["protobufv3-codec"]
```
BusyJay, line 9: I suggest to make "protobuf-codec" default instead just like other crates in the repo.

proto/src/util.rs
```diff
@@ -76,7 +83,11 @@ impl TryFrom<Status> for grpcio::RpcStatus {
 
     fn try_from(value: Status) -> grpcio::Result<Self> {
         #[cfg(feature = "protobuf-codec")]
-        let details = protobuf::Message::write_to_bytes(&value)?;
+        let details = protobuf::Message::write_to_bytes(&value)
+            .map_err(|_err| grpcio::Error::RemoteStopped)?;
+        #[cfg(feature = "protobufv3-codec")]
+        let details = protobufv3::Message::write_to_bytes(&value)
```
BusyJay, line 89: If protobufv3 is re-exported as protobuf, then here only needs to write `protobuf::Message::write_to_bytes`.

interop/src-protobufv3/bin/client.rs
```diff
@@ -0,0 +1,112 @@
+// Copyright 2019 TiKV Project Authors. Licensed under Apache-2.0.
```
BusyJay, line 1: Can interop and benchmark be also merged?

tests-and-examples/Cargo.toml
lb034582341: Good catch, thanks

proto/src/util.rs
lb034582341: Done

interop/src-protobufv3/bin/client.rs
lb034582341: Done, ty

lb034582341:
Done. Since the previous presubmit failed due to linting errors, I took the opportunity to clean this up too. Please let me know if this doesn't belong in this PR

lb034582341:
The last CI error doesn't seem related to my change, and seems fixed on main. Maybe it has been fixed since ?

BusyJay:
Yes, please solve conflicts.

lb034582341:
Done, ty

lb034582341:
OK, CI-wise we should be good (although some tests seem flaky): https://github.com/lb034582341/grpc-rs/actions/runs/5600803677
All these commits are a bit messy. Would you like me to squash them all, or do you prefer to keep the full history ?

BusyJay:
> Would you like me to squash them all

Don't worry, they will be squashed when it's merged.

benchmark/src/worker.rs
```diff
@@ -72,17 +79,26 @@ impl WorkerService for Worker {
...
+            #[cfg(feature = "protobufv3-codec")]
+            let cfg = arg.setup();
             info!("receive client setup: {:?}", cfg);
             let mut client = Client::new(cfg);
             sink.send((ClientStatus::default(), WriteFlags::default()))
                 .await?;
             while let Some(arg) = stream.try_next().await? {
+                #[cfg(feature = "protobuf-codec")]
                 let mark = arg.get_mark();
+                #[cfg(feature = "protobufv3-codec")]
```
BusyJay, line 93: How about using a trait to unify them?

```
trait Accessor {
    type Item;
    fn get(&self) -> &Self::Item;
    fn r#mut(&mut self) -> &mut Self::Item;
}

impl<T> Accessor for Option<T> {
    type Item = T;
    xxx
}

impl<T> Accessor for SinglePtrField<T> {
    type Item = T;
    xxx
}
```
?

lb034582341: Hi, I'm confused on how to do this. My understanding is that you want a generic trait to implement both the types used in v2 and v3, so they can be accessed with a one-liner `.get()` or `.mut()` regardless of the version.
However in this case, the methods have different names `.get_mark()` vs `.mark()` so we'd still need the distinction ? What am I missing ? thanks

BusyJay: So they should be able to be accessed by `.mark.get()`. The accessors `get_xxx` and `xxx` are used to handle the case that `xxx` is missing. It can be implemented in the `trait Accessor` directly.

BusyJay:
@lb034582341 Hi, we may release a new version soon. It would be better if this PR is included to next release. Do you think it can catch the release date (before next week)?

benchmark/src/worker.rs
BusyJay: I see, let's keep it the current way for now.

lb034582341: Sounds good. Then please let me know if you have other comments, I believe all your previous comments are addressed.

lb034582341:
NB: The CI failed due to flakiness, it passed here: https://github.com/lb034582341/grpc-rs/commits/master

compiler/Cargo.toml
```diff
@@ -14,15 +14,18 @@ categories = ["network-programming"]
...
 
 [dependencies]
 protobuf = { version = "2", optional = true }
+protobufv3 = { package = "protobuf", version = "3.2", optional = true }
 prost = { version = "0.11", optional = true }
 prost-build = { version = "0.11", optional = true }
 prost-types = { version = "0.11", optional = true }
 derive-new = { version = "0.5", optional = true }
 tempfile = { version = "3.0", optional = true }
+protobuf-codegen = { version = "3.2.0", optional = true }
```
BusyJay, line 28: Seems not enabled by any features at all.

compiler/src/codegen.rs
```diff
@@ -702,6 +742,41 @@ pub fn gen(
...
 
+#[cfg(feature = "protobufv3-codec")]
+pub fn gen(
+    file_descriptors: &[FileDescriptorProto],
+    files_to_generate: &[String],
+) -> Vec<compiler_plugin::GenResult> {
+    let files_map: HashMap<&str, &FileDescriptorProto> =
+        file_descriptors.iter().map(|f| (f.get_name(), f)).collect();
+
+    let res = protobuf_codegen::Codegen::new()
```
BusyJay, line 753: Why need this? They should be generated by protobuf codegen itself.

health/tests/health_check.rs
```diff
@@ -10,14 +10,42 @@ use std::time::Duration;
...
 #[track_caller]
 fn assert_status(status: ServingStatus, client: &HealthClient, name: &str) {
     let req = HealthCheckRequest {
         service: name.to_string(),
         ..Default::default()
     };
-    let resp = client.check(&req).unwrap();
-    assert_eq!(resp.status, status)
+    let resp: HealthCheckResponse = client.check(&req).unwrap();
+    assert_eq!(response_status_equals(resp, status), true);
```
BusyJay, line 48: Should be `assert!(response_status_equals(resp, status)`.

interop/Cargo.toml
```diff
@@ -4,11 +4,17 @@ version = "0.1.0"
...
+default = ["protobuf-codec"]
+protobuf-codec = ["protobuf", "grpcio/protobuf-codec", "grpcio-proto/protobuf-codec"]
+protobufv3-codec = ["protobufv3", "grpcio/protobufv3-codec", "grpcio-proto/protobufv3-codec"]
+
 [dependencies]
-grpcio = { path = ".." }
-grpcio-sys = { path = "../grpc-sys" }
-grpcio-proto = { path = "../proto" }
-protobuf = "2"
+grpcio = { path = "..", default-features = false}
```
BusyJay, line 13: Should enable boringssl.

proto/src/util.rs
```diff
@@ -75,8 +82,9 @@ impl TryFrom<Status> for grpcio::RpcStatus {
     type Error = grpcio::Error;
 
     fn try_from(value: Status) -> grpcio::Result<Self> {
-        #[cfg(feature = "protobuf-codec")]
-        let details = protobuf::Message::write_to_bytes(&value)?;
+        #[cfg(any(feature = "protobuf-codec", feature = "protobufv3-codec"))]
+        let details = protobuf::Message::write_to_bytes(&value)
+            .map_err(|_err| grpcio::Error::RemoteStopped)?;
```
BusyJay, line 87: Why change it to `RemoteStopped`?

health/tests/health_check.rs
```diff
@@ -10,14 +10,42 @@ use std::time::Duration;
 
 const TEST_SERVICE: &str = "grpc.test.TestService";
 
+#[cfg(feature = "protobuf-codec")]
+mod constants {
+    use crate::proto::ServingStatus;
```
BusyJay, line 15: The code of health module are generated by xtask, maybe you can update xtask instead. And `Serving` is more rusty than `SERVING`.

BusyJay (commented):
Code in proto crate should also be maintained by xtask too.

Rest LGTM

compiler/Cargo.toml
lb034582341: True, this was an oversight. Protobufv3 wasn't enabled at all in the compiler. I had to do a bunch of changes to enable it, please see the next commit.

compiler/src/codegen.rs
lb034582341: Indeed.

health/tests/health_check.rs
lb034582341: Done

interop/Cargo.toml
lb034582341: Done, also for benchmark/ and health/

lb034582341 (commented):
(cancelling review, I missclicked)

proto/src/util.rs
lb034582341: This was to fix a type error, but this doesn't seem needed. I removed both occurrences.

lb034582341:
There's a bit of work to update `xtask`, I need maybe a couple of days. Please let me know if the deadline for the next release is set, I'll try to make it.

lb034582341:
Sorry about messing the history, one commit wasn't signed off and DGO was blocking the CI.
Despite the force push, I didn't change what you already reviewed.

Simple things first, I addressed your 3 comments in https://github.com/tikv/grpc-rs/pull/615/commits/8be9ef547dd7a3fb5405f535f7890add160cba03, https://github.com/tikv/grpc-rs/pull/615/commits/22e7c0ed7b6c274311c5274131c1d7e07b948037 and https://github.com/tikv/grpc-rs/pull/615/commits/610032fc498ed02068f5fc0ac0c25c005b7b0f3a

I added support for v3 in compiler https://github.com/tikv/grpc-rs/pull/615/commits/40c10a8ca5ccc65c3e5a516bfa179f2440befbcb. I removed the previous code that was wrong (was never compiled due to a feature error). Now this simply calls the [rust_protobuf](https://docs.rs/protobuf-codegen/latest/protobuf_codegen/) crate. Finally, maybe we want to discard this commit, and keep `compiler/` mostly about pb2 and prost. `xtask` can call the `rust_protobuf` crate directly. Please let me know what you think.

The comment about xtask required more changes:
- Changed xtask to output pbv3 code https://github.com/tikv/grpc-rs/pull/615/commits/2867cd2bc8511e0466dc34b666fe7bee5800d0b7 .
-- I refactored some code common for v2, v3 and prost into new methods;
-- I didn't manage to find the same plugin mechanism on rust_protobuf so I'm using the compiler/ code to generate the _grpc.rs even for protobuf_v3, this works.
-- I added a renaming `::protobuf::` -> `::protobuf_v3::` to match the imports
-- Nit: Now that we have v2 and v3 code, maybe there's no need for `remove_protobuf_version_constraint()` ? I left it there for now.

Finally, I simplified `health/` in https://github.com/tikv/grpc-rs/pull/615/commits/648b815c4b66cbc6144ba4f326657b8358446331 and regenerated all files in the last commit

compiler/src/codegen.rs
```diff
@@ -702,6 +728,22 @@ pub fn gen(
     results
 }
 
+#[cfg(feature = "protobufv3-codec")]
+pub fn gen(input_files: &[String]) -> anyhow::Result<()> {
+    let res = protobuf_codegen::Codegen::new()
```
BusyJay, line 733: I don't get it. Isn't it just generating the protobuf files? The compiler of grpcio should only be used to generate service definition.

health/Cargo.toml
```diff
@@ -15,12 +15,14 @@ readme = "README.md"
...
 default = ["protobuf-codec"]
 protobuf-codec = ["grpcio/protobuf-codec", "protobuf"]
+protobufv3-codec = ["grpcio/protobufv3-codec", "protobufv3"]
 prost-codec = ["grpcio/prost-codec", "prost"]
 
 [dependencies]
 futures-executor = "0.3"
 futures-util = { version = "0.3", default-features = false, features = ["std"] }
-grpcio = { path = "..", version = "0.12.1", default-features = false }
+grpcio = { path = "..", version = "0.12.1", default-features = false, features = ["boringssl"]}
```
BusyJay, line 24: No, users should decide which ssl to use. `health` is a public crate that is different from benchmark and interop.

compiler/Cargo.toml
```diff
@@ -14,15 +14,19 @@ categories = ["network-programming"]
 [features]
 default = ["protobuf-codec"]
 protobuf-codec = ["protobuf"]
+protobufv3-codec = ["protobufv3", "protobuf-codegen", "anyhow"]
```
BusyJay, line 17: If the generated service code of protobuf-v3 is exact the same as protobuf-v2, then it seems no need to add a standalone features.

health/Cargo.toml
lb034582341: Reverted, thanks

compiler/src/codegen.rs
lb034582341: I got confused, thinking it should generate both. Removed. Thanks

BusyJay (commented):
Rest LGTM

health/tests/health_check.rs
```diff
@@ -10,14 +10,40 @@ use std::time::Duration;
...
 const TEST_SERVICE: &str = "grpc.test.TestService";
 
+#[cfg(feature = "protobuf-codec")]
+use crate::proto::ServingStatus;
+
+#[cfg(feature = "protobufv3-codec")]
+use grpcio_health::ServingStatus;
+
+/*
+#[cfg(feature = "protobufv3-codec")]
```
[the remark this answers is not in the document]
lb034582341: My bad.

health/tests/health_check.rs
```diff
@@ -124,12 +140,16 @@ fn test_health_watch() {
...
     let mut seen = 0;
     loop {
-        let resp = block_on(statuses.next()).unwrap().unwrap();
-        if resp.status != ServingStatus::Unknown {
-            seen += 1;
-            continue;
+        match block_on(statuses.next()).unwrap() {
+            Err(e) => panic!("unexpected error {:?}", e),
+            Ok(r) => {
+                if response_status_equals(r, ServingStatus::Unknown) {
```
BusyJay, line 146: You mean `if !response_status_equals(r, ServingStatus::Unknown)`?

lb034582341:
I've included your changes, thanks for the review. CI passed here https://github.com/lb034582341/grpc-rs/actions/runs/5796423100

BusyJay (approved):

BusyJay:
Thank you!

lb034582341:
>@lb034582341 Hi, we may release a new version soon. It would be better if this PR is included to next release. Do you think it can catch the release date (before next week)?

Hi @BusyJay, is there anything I can do to help with the release ? thanks!

BusyJay:
It was already released in crates.io. I just forgot to update github release page. And the page was just updated.

lb034582341:
Ah, right. Thanks a lot!

Eclextic:
I would like to ask how this works with programmatic code generation because as far as I can see, there is nothing indicating that that works.

Am I correct? Would one have to use manual generation?

PR #622, opened 2023-07-12
gate auth context under `_secure`

tabokie:
Windows build is failing without ssl because `AuthContext` uses methods in "grpc_security.h". See https://github.com/tikv/raft-engine/actions/runs/5493202302/jobs/10013725416 for exact build errors.


BusyJay:
Interesting. Though auth_context related symbols should also in grpc_unsecure library. https://github.com/grpc/grpc/blob/v1.56.x/CMakeLists.txt#L2914

BusyJay (approved):

PR #623, opened 2023-07-19
Fix unknown log file and module

Connor1996:
Logs are always with unknown file name when redirecting logs to `slog`, see
截屏2023-07-18 17 56 09

It's because slog would only take static str into account
https://github.com/slog-rs/stdlog/blob/252b6bb6435a8a3ba44748b2e1626aa676508097/lib.rs#L73

With this fix, the file name is not unknown anymore
截屏2023-07-18 17 55 39


tonyxuqqi (approved):

src/log_util.rs
```diff
@@ -30,9 +30,9 @@ extern "C" fn delegate(c_args: *mut gpr_log_func_args) {
         &Record::builder()
             .args(format_args!("{msg}"))
             .level(level)
-            .file(file_str.into())
+            .file_static(file_str.into())
```
BusyJay, line 33: Is it really static?

Connor1996: yes, file name is generated by macro, see https://github.com/tikv/grpc/blob/996605a5e62f3f00043ac8d3ebca84523bc2dd76/include/grpc/impl/codegen/log.h#L55

BusyJay (approved):

Issue #627, opened 2023-08-03
0.12.1+1.56.2-patched failed to link systemd


fortime:
**Describe the bug**
I am building an executable with grpc-rs ([this commit](https://github.com/tikv/grpc-rs/commit/90dad054eab3a94baf734203bbb7910d8668225a)) on Arch Linux. Linking is failed because of `-lsystemd` is not provided.
Error log:
```
/usr/bin/ld: /run/user/1000/cargo-target/debug/deps/libgrpcio_sys-95ecc2781c08cbb9.rlib(systemd_utils.cc.o): in function `set_matching_sd_unix_fd(grpc_tcp_server*, grpc_resolved_address const*, int, int)':
/home/fortime/.cargo/git/checkouts/grpc-rs-f94f0a462dfbe83f/339c325/grpc-sys/grpc/src/core/lib/iomgr/systemd_utils.cc:40: undefined reference to `sd_is_socket_unix'
/usr/bin/ld: /run/user/1000/cargo-target/debug/deps/libgrpcio_sys-95ecc2781c08cbb9.rlib(systemd_utils.cc.o): in function `set_matching_sd_inet_fd(grpc_tcp_server*, grpc_resolved_address const*, int, int, int, int)':
/home/fortime/.cargo/git/checkouts/grpc-rs-f94f0a462dfbe83f/339c325/grpc-sys/grpc/src/core/lib/iomgr/systemd_utils.cc:53: undefined reference to `sd_is_socket_inet'
/usr/bin/ld: /home/fortime/.cargo/git/checkouts/grpc-rs-f94f0a462dfbe83f/339c325/grpc-sys/grpc/src/core/lib/iomgr/systemd_utils.cc:54: undefined reference to `sd_is_socket_sockaddr'
/usr/bin/ld: /run/user/1000/cargo-target/debug/deps/libgrpcio_sys-95ecc2781c08cbb9.rlib(systemd_utils.cc.o): in function `set_matching_sd_fds(grpc_tcp_server*, grpc_resolved_address const*, int)':
/home/fortime/.cargo/git/checkouts/grpc-rs-f94f0a462dfbe83f/339c325/grpc-sys/grpc/src/core/lib/iomgr/systemd_utils.cc:69: undefined reference to `sd_listen_fds'
collect2: error: ld returned 1 exit status

some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
use the `-l` flag to specify native libraries to link
[...]

PR #629, opened 2023-08-07
fix link error on os with systemd

fortime:
fix: [#627](https://github.com/tikv/grpc-rs/issues/627)

search systemd path from `CMakeCache.txt`, and output it to cargo.


grpc-sys/build.rs
```diff
@@ -293,9 +293,46 @@ fn build_grpc(cc: &mut cc::Build, library: &str) {
         }
     }
 
+    // systemd is only added when platform is linux or posix(not linux, darwin, ios, android)
+    // in grpc CMakeLists.txt
+    if get_env("CARGO_CFG_TARGET_FAMILY").map_or(false, |s| s == "unix")
+        && get_env("CARGO_CFG_TARGET_OS").map_or(false, |s| {
+            s == "linux" || !(s == "macos" || s == "ios" || s == "android")
```
BusyJay, line 300: How about always searching cache and link if found?

BusyJay (approved):

PR #641, opened 2023-11-29
Add methods taking file descriptor as OwnedFd.

qwandor:
The `OwnedFd` type confers ownership of the file descriptor, so these can be safe methods rather than relying on the caller not to use the file descriptor anywhere else after making the call.


BusyJay (approved):

qwandor:
Thanks for the review! Can you merge the PR too? The CI failures are all existing issues as far as I can see, not changed by this PR.

BusyJay:
Thanks!

Issue #620, opened 2023-07-06
Failed to build on windows


MichaelScofield:
**Describe the bug**

Unable to build grpc-rs on the latest windows 11. Here's the error logs: (from cargo build)

```txt
PS D:\repo\grpc-rs\grpc-sys> cargo build
warning: patch for `grpcio-compiler` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
   Compiling log v0.4.19
   Compiling env_logger v0.9.3
   Compiling bindgen v0.59.2
   Compiling grpcio-sys v0.12.1+1.46.5-patched (D:\repo\grpc-rs\grpc-sys)
error: failed to run custom build command for `grpcio-sys v0.12.1+1.46.5-patched (D:\repo\grpc-rs\grpc-sys)`

Caused by:
  process didn't exit successfully: `D:\repo\grpc-rs\target\debug\build\grpcio-sys-f7652c89275eba94\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-changed=grpc_wrap.cc
  cargo:rerun-if-changed=grpc
  cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS
  cargo:rerun-if-env-changed=GRPCIO_SYS_USE_PKG_CONFIG
  cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS
  cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS
  cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS
  cargo:rerun-if-env-changed=CXX
  OPT_LEVEL = Some("0")
  TARGET = Some("x86_64-pc-windows-msvc")
  HOST = Some("x86_64-pc-windows-msvc")
  cargo:rerun-if-env-changed=CC_x86_64-pc-windows-msvc
  CC_x86_64-pc-windows-msvc = None
  cargo:rerun-if-env-changed=CC_x86_64_pc_windows_msvc
  CC_x86_64_pc_windows_msvc = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64-pc-windows-msvc
  CFLAGS_x86_64-pc-windows-msvc = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64_pc_windows_msvc
[...]

PR #644, opened 2024-01-02
grpc-sys: Fix build failure on uncommon linux build targets

harryoooooooooo:
Hi owners, I'm Hsin-chen from Google. Recently we stepped into a build failure in our product, below is the failure reason.

After some investigation I believe the culprit is around the [`config_binding_path`](https://github.com/tikv/grpc-rs/blob/5442991f322901d28e318bac5736abe13a10794c/grpc-sys/build.rs#L483) function: Our build target is not listed as a supported platform but the bindings are not generated either because our target_os is linux. Eventually we're only able to build with `feature = "_gen-bindings"`.

I've attached a possible fix. Could you please help take a look? Thanks!

```
floss-9999: error: couldn't read /build/brya/var/cache/portage/net-wireless/floss/cros-rust/x86_64-cros-linux-gnu/release/build/grpcio-sys-947c4c4fa7b189ac/out/grpc-bindings.rs: No such file or directory (os error 2)
floss-9999:  --> [REGISTRY]/grpcio-sys-0.13.0+1.56.2-patched/src/lib.rs:8:5
floss-9999:   |
floss-9999: 8 |     include!(env!("BINDING_PATH"));
floss-9999:   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
floss-9999:   |
floss-9999:   = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)
floss-9999:
floss-9999: error: could not compile `grpcio-sys` (lib) due to previous error
```

### DIFF
--- a/grpc-sys/build.rs
+++ b/grpc-sys/build.rs
@@ -481,24 +481,24 @@ fn bindgen_grpc(file_path: &Path) {
 // need to be updated by default unless the _gen-bindings feature is specified.
 // Other platforms use bindgen to generate the bindings every time.
 fn config_binding_path() {
-    let target = env::var("TARGET").unwrap();
-    let file_path: PathBuf = match target.as_str() {
-        "x86_64-unknown-linux-gnu"
-        | "x86_64-unknown-linux-musl"
-        | "aarch64-unknown-linux-musl"
-        | "aarch64-unknown-linux-gnu"
-        | "x86_64-apple-darwin"
-        | "aarch64-apple-darwin" => {
-            // Cargo treats nonexistent files changed, so we only emit the rerun-if-changed
-            // directive when we expect the target-specific pre-generated binding file to be
-            // present.
-            println!("cargo:rerun-if-changed=bindings/bindings.rs");
+    let file_path: PathBuf;
 
-            PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
-                .join("bindings")
-                .join("bindings.rs")
-        }
-        _ => PathBuf::from(env::var("OUT_DIR").unwrap()).join("grpc-bindings.rs"),
+    #[cfg(not(any(
+        feature = "_gen-bindings",
+        not(all(
+            any(target_os = "linux", target_os = "macos"),
+            any(target_arch = "x86_64", target_arch = "aarch64")
+        ))
+    )))]
+    {
+        // Cargo treats nonexistent files changed, so we only emit the rerun-if-changed
+        // directive when we expect the target-specific pre-generated binding file to be
+        // present.
+        println!("cargo:rerun-if-changed=bindings/bindings.rs");
+
+        file_path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
+            .join("bindings")
+            .join("bindings.rs")
     };
 
     #[cfg(any(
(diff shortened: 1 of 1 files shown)


grpc-sys/build.rs
```diff
@@ -481,24 +481,24 @@ fn bindgen_grpc(file_path: &Path) {
...
-            println!("cargo:rerun-if-changed=bindings/bindings.rs");
-
-            PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap())
-                .join("bindings")
-                .join("bindings.rs")
-        }
-        _ => PathBuf::from(env::var("OUT_DIR").unwrap()).join("grpc-bindings.rs"),
+    let file_path: PathBuf;
+
+    #[cfg(not(any(
```
BusyJay, line 486: Instead of changing here, you should make the conditional check at L505~508 more accurate. And Cargo.toml should also be updated to include necessary dependencies.

BusyJay: What I mean is bindings is pregenerated only on following targets:
```
 "x86_64-unknown-linux-gnu"
        | "x86_64-unknown-linux-musl"
        | "aarch64-unknown-linux-musl"
        | "aarch64-unknown-linux-gnu"
        | "x86_64-apple-darwin"
        | "aarch64-apple-darwin"
```
So what we need to do is figure out a more precise `#[cfg(xxx)]` that exactly matches those targets.

harryoooooooooo: Is it necessary to approach this in compile time, i.e., with `#[cfg(xxx)]` tag?

If not then I think my proposal already addressed this by `cfg!(feature = "_gen-bindings")`.

BusyJay: > Is it necessary to approach this in compile time, i.e., with #[cfg(xxx)] tag?

Yes. The point of pre-generating bindings is to avoid downloading bindgen and depends on LLVM on common platforms.

Eclextic:
Can we please finally merge this?

BusyJay:
@Eclextic what issue did you meet? Master branch has already changed the target match rules, so I think for uncommon targets, bindings should be generated at compile time already.

harryoooooooooo:
I don't see any revalent changes after this PR so I believe the issue is still there. An uncommon linux target would enter neither of the checks below, and as a result there's no bindings.rs file.

In my opinion the two checks must be complement (like what this PR does) to cover all cases. As per #discussion_r1602818903 the first one seems to be preferred, so I guess we shall chage the second one correspondingly.
```
    let file_path: PathBuf = match target.as_str() {
        "x86_64-unknown-linux-gnu"
        | "x86_64-unknown-linux-musl"
        | "aarch64-unknown-linux-musl"
        | "aarch64-unknown-linux-gnu"
        | "x86_64-apple-darwin"
        | "aarch64-apple-darwin" =>
```

```
    #[cfg(any(
        feature = "_gen-bindings",
        not(all(
            any(target_os = "linux", target_os = "macos"),
            any(target_arch = "x86_64", target_arch = "aarch64")
        ))
    ))]
```

Eclextic:
> @Eclextic what issue did you meet? Master branch has already changed the target match rules, so I think for uncommon targets, bindings should be generated at compile time already.

I mean I am gonna be honest. In hindsight I am kind of unsure if this is related at all, but whenever one cargo cleans and recompiles the project I get a cmake deprecation error on linux. I think this is technically unrelated to this issue and moreover related to #620.

So yeah I am sorry. I'll continue posting there.

Issue #659, opened 2025-05-16
Compilation error: GPR_ASSERT not found


nikurt:
**Describe the bug**
When building the `grpcio-sys` crate (version 0.13.0+1.56.2-patched, as well as branch master at the latest commit `0b1798dc`) with `cargo build`, the compilation fails due to undefined `GPR_ASSERT` macros in `grpc_wrap.cc`.

**To Reproduce**
1. `gh repo clone tikv/grpc-rs`
2. `cd grpc-rs`
3. `cargo build`
4. Observe the compilation failed

**Expected behavior**
The crate builds

**System information**
* CPU architecture: x86_64
* Distribution and kernel version: Arch Linux 6.14.6-arch1-1
* SELinux on?: No
* Any other system details we should know?:
* * `cargo 1.86.0 (adf9b6ad1 2025-02-28)`
* * `gcc (GCC) 15.1.1 20250425`

**Additional context**
The compilation error message:
```
  cargo:warning=grpc_wrap.cc:175:3: error: use of undeclared identifier 'GPR_ASSERT'
  cargo:warning=  175 |   GPR_ASSERT(array->count <= array->capacity);
  cargo:warning=      |   ^
  cargo:warning=grpc_wrap.cc:191:3: error: use of undeclared identifier 'GPR_ASSERT'
  cargo:warning=  191 |   GPR_ASSERT(index < array->count);
  cargo:warning=      |   ^
  cargo:warning=grpc_wrap.cc:198:3: error: use of undeclared identifier 'GPR_ASSERT'
  cargo:warning=  198 |   GPR_ASSERT(index < array->count);
  cargo:warning=      |   ^
  cargo:warning=grpc_wrap.cc:210:3: error: use of undeclared identifier 'GPR_ASSERT'
  cargo:warning=  210 |   GPR_ASSERT(array->count <= array->capacity);
  cargo:warning=      |   ^
  cargo:warning=grpc_wrap.cc:438:3: error: use of undeclared identifier 'GPR_ASSERT'
  cargo:warning=  438 |   GPR_ASSERT(args);
  cargo:warning=      |   ^
  cargo:warning=grpc_wrap.cc:439:3: error: use of undeclared identifier 'GPR_ASSERT'
  cargo:warning=  439 |   GPR_ASSERT(index < args->num_args);
  cargo:warning=      |   ^
  cargo:warning=grpc_wrap.cc:447:3: error: use of undeclared identifier 'GPR_ASSERT'
[...]

PR #660, opened 2025-05-16
Fix: Define GPR_ASSERT macro for compatibility

nikurt:
Added a fallback definition for GPR_ASSERT using <assert.h> to handle cases where it is not predefined. This ensures smoother compilation and avoids potential issues with missing assertions.

Fixes https://github.com/tikv/grpc-rs/issues/659


grpc-sys/grpc_wrap.cc
```diff
@@ -66,6 +66,11 @@
 #define GPR_CALLTYPE
 #endif
 
+#ifndef GPR_ASSERT
+#include <assert.h>
```
BusyJay, line 70: How about switching to absl check just like what upstream did?

nikurt: Thanks for the suggestion, haven't thought of that.

Observations:

* `absl/log/absl_check.h` has a static assert that the C++ variant is at least C++14
* `ABSL_CHECK` depends on `using string_view = std::string_view` which is only available in C++17.

Attempted a prototype https://github.com/tikv/grpc-rs/pull/661 which upgrades C++ to C++14, but can't make it compile.

A different prototype https://github.com/tikv/grpc-rs/pull/662 upgrades C++ to C++17 and it compiles.

Here is what happens with a C++14 upgrade:

```
% cargo build
   Compiling grpcio-sys v0.13.0+1.56.2-patched (/home/nkurtov/code/grpc-rs/grpc-sys)
warning: grpcio-sys@0.13.0+1.56.2-patched: In file included from /usr/include/absl/log/internal/nullstream.h:37,
warning: grpcio-sys@0.13.0+1.56.2-patched:                  from /usr/include/absl/log/internal/check_op.h:40,
warning: grpcio-sys@0.13.0+1.56.2-patched:                  from /usr/include/absl/log/internal/check_impl.h:19,
warning: grpcio-sys@0.13.0+1.56.2-patched:                  from /usr/include/absl/log/absl_check.h:38,
warning: grpcio-sys@0.13.0+1.56.2-patched:                  from grpc_wrap.cc:71:
warning: grpcio-sys@0.13.0+1.56.2-patched: /usr/include/absl/strings/string_view.h:53:26: error: 'string_view' in namespace 'std' does not name a type
warning: grpcio-sys@0.13.0+1.56.2-patched:    53 | using string_view = std::string_view;
warning: grpcio-sys@0.13.0+1.56.2-patched:       |                          ^~~~~~~~~~~
warning: grpcio-sys@0.13.0+1.56.2-patched: /usr/include/absl/strings/string_view.h:53:21: note: 'std::string_view' is only available from C++17 onwards
warning: grpcio-sys@0.13.0+1.56.2-patched:    53 | using string_view = std::string_view;
warning: grpcio-sys@0.13.0+1.56.2-patched:       |                     ^~~
```

BusyJay: I see. Upgrading C++ standard will introduce breaking change, let's keep it this way for now.

BusyJay (approved):
Thank you!

Issue #664, opened 2025-07-26
benchmark: build fail and test fail.


Dog-Du:
**Describe the bug**

**To Reproduce**
1. Go to 'benchmark'
2. Run 'cargo build'
3. See error
**Expected behavior**
compile and no error:

**Additional context**
## build fail
```
❯ cargo build
   Compiling benchmark v0.1.0 (~/grpc-rs/benchmark)
error: cannot find derive macro `Parser` in this scope
  --> benchmark/src/main.rs:23:10
   |
23 | #[derive(Parser)]
   |          ^^^^^^
   |
note: `Parser` is imported here, but it is only a trait, without a derive macro
  --> benchmark/src/main.rs:12:5
   |
12 | use clap::Parser;
   |     ^^^^^^^^^^^^

error: cannot find attribute `arg` in this scope
  --> benchmark/src/main.rs:26:7
   |
26 |     #[arg(long)]
   |       ^^^

error[E0599]: no function or associated item named `parse` found for struct `WorkerCli` in the current scope
  --> benchmark/src/main.rs:31:26
   |
24 | struct WorkerCli {
   | ---------------- function or associated item `parse` not found for this struct
...
31 |     let cli = WorkerCli::parse();
   |                          ^^^^^ function or associated item not found in `WorkerCli`
   |
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following traits define an item `parse`, perhaps you need to implement one of them:
           candidate #1: `Parser`
           candidate #2: `TypedValueParser`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `benchmark` (bin "qps_worker") due to 3 previous errors
```
we need to add `derive` feature when importing `clag`

## test fail
```
++ dirname tools/run_tests/performance/run_worker_rust.sh
+ cd tools/run_tests/performance/../../..
+ ../grpc-rs/target/release/qps_worker --driver_port=10110
error: unexpected argument '--driver_port' found

  tip: a similar argument exists: '--driver-port'

Usage: qps_worker <DRIVER_PORT0|--driver-port <driver-port>>

For more information, try '--help'.
```
we need to support the `--driver_port`

PR #665, opened 2025-07-26
fix benchmark build fail and test fail #664

Dog-Du:
close issue: #664 
build benchmark fail.
we need to add `derive` feature when importing `clap` 
benchmark test fail.
we need to add the argument `--driver_port`

### DIFF
--- a/benchmark/src/main.rs
+++ b/benchmark/src/main.rs
@@ -33,7 +33,7 @@ struct WorkerCli {
 
 fn main() {
     let cli = WorkerCli::parse();
-    let port = cli.driver_port0.unwrap_or(cli.driver_port1.unwrap_or(32766));
+    let port = cli.driver_port0.unwrap_or(cli.driver_port1.unwrap_or(8080));
 
     let _log_guard = init_log(
         env::var(LOG_FILE)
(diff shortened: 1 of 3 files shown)


benchmark/src/main.rs
```diff
@@ -21,15 +21,19 @@ const LOG_FILE: &str = "GRPCIO_BENCHMARK_LOG_FILE";
...
+#[group(required = true, multiple = false)]
 struct WorkerCli {
     /// The port the worker should listen on. For example, 8080
-    #[arg(long)]
-    driver_port: Option<u16>,
+    #[arg(id = "driver_port", long)]
+    driver_port0: Option<u16>,
+    /// The port the worker should listen on. For example, 8080
+    #[arg(id = "driver-port", long)]
+    driver_port1: Option<u16>,
```
overvenus, line 31: Why add a new port? what's difference between `driver_port` and `driver-port`?

Dog-Du: No difference, only one of them is not `None`, just to support both `--driver-port` and `--driver_port`. 
we need `--driver_port` when running `python3 tools/run_tests/run_performance_tests.py -l rust`

